PR #4, opened 2023-03-21
CostControlBandit: Take action with highest proba in case several actions have same cost

spiaz:


tests/test_strategy.py
```diff
@@ -198,26 +198,28 @@ def test_select_action_cc(a_list_str, a_list_float):
...
-    assert "a1" == c.select_action(p=p, actions=actions)
+    # if cost factor is 1 => return the action with min cost and best proba
+    assert "a4" == c.select_action(p=p, actions=actions)
 
     # if cost factor is 0 => return the action with highest p (classic bandit)
     c.set_subsidy_factor(subsidy_factor=0)
     assert "a2" == c.select_action(p=p, actions=actions)
 
-    # otherwise, return the cheapest action that is still feasible
+    # otherwise, return the cheapest feasible action with best probability
```
dariodandrea, line 220: Suggestion for naming convention:
`# otherwise, return the cheapest feasible action with the highest probability `

tests/test_strategy.py
```diff
@@ -198,26 +198,28 @@ def test_select_action_cc(a_list_str, a_list_float):
...
         "a2": BetaCC(cost=actions_cost["a2"]),
         "a3": BetaCC(cost=actions_cost["a3"]),
+        "a4": BetaCC(cost=actions_cost["a4"]),
+        "a5": BetaCC(cost=actions_cost["a5"]),
     }
 
     c = CostControlBandit(subsidy_factor=1)
-    # if cost factor is 1 => return the action with min cost
-    assert "a1" == c.select_action(p=p, actions=actions)
+    # if cost factor is 1 => return the action with min cost and best proba
```
dariodandrea, line 213: Suggestion for the naming convention:
` # if cost factor is 1 => return the action with min cost and the highest sampled probability `

pybandits/strategy.py
```diff
@@ -221,11 +221,14 @@ def select_action(self, p: Dict[ActionId, Probability], actions: Dict[ActionId,
         # define the set of feasible actions
         feasible_actions = [a for a in p.keys() if p[a] >= (1 - self.subsidy_factor) * max_p]
 
-        # feasible action costs
-        _, cheapest_action = min((b.cost, a) for a, b in actions.items() if a in feasible_actions)
+        # feasible actions and their characteristics
+        sortable_actions = [(actions[a].cost, 1 - p[a], a) for a in feasible_actions]
+
+        # best action is the one with cheapest cost (+ higher proba in case of equality)
+        _, _, best_action = sorted(sortable_actions)[0]
```
dariodandrea, line 228: To use the same naming convention as other strategies, I would suggest renaming `best_action` to `selected_action`

```
# select the action with the cheapest cost (and the highest probability in case of cost equality)
 _, _, selected_action = sorted(sortable_actions)[0]
```

pybandits/strategy.py
```diff
@@ -221,11 +221,14 @@ def select_action(self, p: Dict[ActionId, Probability], actions: Dict[ActionId,
         # define the set of feasible actions
         feasible_actions = [a for a in p.keys() if p[a] >= (1 - self.subsidy_factor) * max_p]
 
-        # feasible action costs
-        _, cheapest_action = min((b.cost, a) for a, b in actions.items() if a in feasible_actions)
+        # feasible actions and their characteristics
```
dariodandrea, line 224: Small suggestion:

`# feasible actions enriched with their characteristics (cost, 1-probability, action_id)`

tests/test_strategy.py
j3rom3c: I would also be more precise:
# otherwise, return the cheapest feasible action. In case of more than one action, the action with the highest sampled probability is returned

tests/test_strategy.py
j3rom3c: If cost factor is 1, then there is no cost consideration and the action with the highest sampled probability is returned. Indeed, in such a case the set of feasible action is made of one action, the one with the highest expected reward

j3rom3c:
few suggestions on "comments"

tests/test_strategy.py
dariodandrea: Also, be careful there is a legacy typo `cost factor` should be `subsidy_factor `

spiaz: I think Dario's version is enough, since we are speaking about a test here 😉

spiaz:
Suggestion implemented, thank you for the feedback!

fsimond (approved):

pybandits/strategy.py
```diff
@@ -221,11 +221,14 @@ def select_action(self, p: Dict[ActionId, Probability], actions: Dict[ActionId,
         # define the set of feasible actions
         feasible_actions = [a for a in p.keys() if p[a] >= (1 - self.subsidy_factor) * max_p]
 
-        # feasible action costs
-        _, cheapest_action = min((b.cost, a) for a, b in actions.items() if a in feasible_actions)
+        # feasible actions enriched with their characteristics (cost, 1-probability, action_id)
+        sortable_actions = [(actions[a].cost, 1 - p[a], a) for a in feasible_actions]
+
+        # best action is the one with cheapest cost (+ higher proba in case of equality)
```
dariodandrea, line 227: I guess we can also modify the comment here:
`# select the action with the cheapest cost (and the highest probability in case of cost equality)`

spiaz: Done! thanks

dariodandrea (approved):
Great job @spiaz! LGTM :)

PR #6, opened 2023-04-25
fix tests

spiaz:
### CHANGES
- fix failing tests: consider probability in case of equal cost of actions

### DIFF
--- a/tests/test_strategy.py
+++ b/tests/test_strategy.py
@@ -58,11 +58,6 @@ def test_select_action_classic_bandit(a_list_str, a_list_float):
     assert max(p, key=p.get) == c.select_action(p=p)
 
 
-# def test_is_compatible():
-#     assert ClassicBandit().is_compatible_with_model(Beta())
-# assert not ClassicBandit().is_compatible_with_model(BetaCC(cost=1))
-
-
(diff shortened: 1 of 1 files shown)


tests/test_strategy.py
```diff
@@ -242,18 +244,20 @@ def test_select_action_logic_corner_cases(a_list_p, a_list_cost):
     }
 
     c = CostControlBandit(subsidy_factor=1)
-    # if cost factor is 1 => return the action with min cost
-    assert min(actions_cost, key=actions_cost.get) == c.select_action(p=p, actions=actions)
+    # if cost factor is 1 => return the action with min cost and max probability in case
```
dariodandrea, line 247: Suggestion for this comment (as it is in the code):

`# select the action with the cheapest cost (and the highest probability in case of cost equality)`

dariodandrea:
Just a minor comment, otherwise LGTM @spiaz :)

dariodandrea:
You can also delete all the commented `test_is_compatible()` in `test_strategy.py` if they are not relevant anymore.

spiaz:
Ok applying the changes and merging. 
Thanks for the code review @dariodandrea 🙌

PR #7, opened 2023-04-25
fix test_select_action_logic_corner_cases

spiaz:
## CHANGES
- fix `test_select_action_logic_corner_cases` and `CostControlBandit.select_action` logic

While fixing this test I found a nice float point approximation bug in the code 🙃


dariodandrea:
Great catch! :)

dariodandrea (approved):

PR #10, opened 2023-05-03
Add Bayesian Logistic Regression model

dariodandrea:
* Add class StudentT
* Add class BaseBayesianLogisticRegression
* Add class BayesianLogisticRegression
* Add unit tests accordingly

### DIFF
--- a/pybandits/model.py
+++ b/pybandits/model.py
@@ -22,12 +22,27 @@
 
 
 from random import betavariate
-from typing import List
+from typing import List, Tuple
 
-from numpy import sqrt
-from pydantic import NonNegativeFloat, PositiveInt, root_validator, validate_arguments
+from numpy import array, c_, exp, insert, mean, multiply, ones, sqrt, std
+from numpy.typing import ArrayLike
+from pydantic import (
+    Field,
+    NonNegativeFloat,
+    PositiveInt,
+    confloat,
+    root_validator,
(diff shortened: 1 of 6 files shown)


pybandits/model.py
```diff
@@ -186,3 +201,208 @@ class BetaMOCC(BaseBetaMO):
...
+
+            context = Data("context", context)
+            rewards = Data("rewards", rewards)
+
+            # Likelihood (sampling distribution) of observations
+            weighted_sum = Deterministic("weighted_sum", alpha + dot(betas, context.T))
+            p = Deterministic("p", sigmoid(weighted_sum))
+
+            # Bernoulli random vector with probability of success given by sigmoid function and actual data as observed
+            _ = Bernoulli("likelihood", p=p, observed=rewards)
```
spiaz, line 363: I guess it's normal in pymc, but how strange this syntax is! the Bernoulli class gets instantiated and isn't used anywhere 🤷‍♂️🤯

spiaz (approved):
Great contribution, and good job @dariodandrea!! 🫶

pybandits/model.py
dariodandrea: Yes, indeed this is how Pymc works. You define the whole model structure and then everything is used implicitly in the sample() function when you update the traces.

PR #17, opened 2023-10-16
feature/fix-style-with-pre-commit-all-files

dariodandrea:
* Fix style in docs files with pre-commit
* Fix style in all python files with pre-commit


spiaz (approved):
🔝

PR #32, opened 2024-07-28
Epsilon Greedy Strategy Wrapper

shaharbar1:
Change log:
 1. Added pytest_mock for mocking pytests in pyproject.toml
 2. Added _make_epsilon_greedy functionality as a static method of Strategy in base.py. The method wraps the native select_action of the strategy with epsilon_greedy approach.
 3. Added epsilon and default_action to all smab.py and cmab.py classes and cold start methods
 4. Added test suite for the epsilon greedy functionlity

### DIFF
--- a/pybandits/base.py
+++ b/pybandits/base.py
@@ -24,6 +24,7 @@
 from abc import ABC, abstractmethod
 from typing import Any, Dict, List, NewType, Optional, Set, Tuple, Union
 
+import numpy as np
 from pydantic import (
     BaseModel,
     Extra,
@@ -31,6 +32,7 @@ from pydantic import (
     confloat,
     conint,
     constr,
+    root_validator,
     validate_arguments,
     validator,
 )
@@ -91,10 +93,14 @@ class BaseMab(PyBanditsBaseModel, ABC):
         The list of possible actions, and their associated Model.
     strategy: Strategy
         The strategy used to select actions.
+    epsilon: Optional[Float01], defaults to None
+        The probability of selecting a random action.
     """
 
     actions: Dict[ActionId, Model]
     strategy: Strategy
+    epsilon: Optional[Float01]
+    default_action: Optional[ActionId]
 
     @validator("actions", pre=True)
     @classmethod
@@ -102,6 +108,14 @@ class BaseMab(PyBanditsBaseModel, ABC):
         if len(v) < 2:
             raise AttributeError("At least 2 actions should be defined.")
         return v
+
+    @root_validator
+    def check_default_action(cls, values):
+        if not values["epsilon"] and values["default_action"]:
+            raise AttributeError("A default action should only be defined when epsilon is defined.")
+        if values["default_action"] and values["default_action"] not in values["actions"]:
+            raise AttributeError("The default action should be defined in the actions.")
+        return values
 
     def _get_valid_actions(self, forbidden_actions: Optional[Set[ActionId]]) -> Set[ActionId]:
         """
@@ -125,6 +139,8 @@ class BaseMab(PyBanditsBaseModel, ABC):
         valid_actions = set(self.actions.keys()) - forbidden_actions
         if len(valid_actions) == 0:
             raise ValueError("All actions are forbidden. You must allow at least 1 action.")
+        if self.default_action and self.default_action not in valid_actions:
+            raise ValueError("The default action is forbidden.")
 
         return valid_actions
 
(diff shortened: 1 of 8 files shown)


pybandits/base.py
```diff
@@ -74,6 +75,47 @@ class Strategy(PyBanditsBaseModel, ABC):
...
+    def __init__(self, **data: Any) -> None:
+        super().__init__(**data)
+        if self.epsilon is not None:
+            self.select_action = self._make_epsilon_greedy(self.epsilon, self.default_action, self.select_action)
+
+    @staticmethod
+    @validate_arguments
+    def _make_epsilon_greedy(epsilon: Float01, default_action: Optional[ActionId], select_action: Callable):
+        """
+        Wraps a select_action function with epsilon-greedy strategy.
```
dariodandrea, line 90: Can you extend the docstring specifying what is this function doing? 
I mean the logic: `return default_action if default_action else np.random.choice(list(p.keys()))`

tests/test_epsilon_greedy.py
```diff
@@ -0,0 +1,205 @@
...
+    SmabBernoulli,
+    SmabBernoulliBAI,
+    SmabBernoulliCC,
+    create_smab_bernoulli_mo_cc_cold_start,
+    create_smab_bernoulli_mo_cold_start,
+)
+from pybandits.strategy import ClassicBandit
+
+
+########################################################################################################################
```
dariodandrea, line 48: Please move this part in the file test_base.py

pybandits/base.py
```diff
@@ -201,3 +206,45 @@ def get_state(self) -> (str, dict):
         model_name = self.__class__.__name__
         state: dict = self.dict()
         return model_name, state
+
+    @validate_arguments
+    def _select_action(
```
dariodandrea, line 211: Let's call this function with an "epsilon-greedy" related name. E.g. `_select_action_epsilon_greedy`

pybandits/base.py
```diff
@@ -201,3 +206,45 @@ def get_state(self) -> (str, dict):
...
+    def _select_epsilon_greedy_action(
+        self,
+        p: Union[Dict[ActionId, float], Dict[ActionId, Probability], Dict[ActionId, List[Probability]]],
+        actions: Optional[Dict[ActionId, Model]] = None,
+    ) -> ActionId:
+        """
+        Wraps self.strategy.select_action function with epsilon-greedy strategy,
+        such that with probability epsilon a default_action is selected,
+        and with probability 1-epsilon the select_action function is triggered to choose action.
+        If no default_action is provided, a random action is selected.
```
dariodandrea, line 220: Please add the reference paper of the epsilon greedy strategy (as it's done in strategy.py)

shaharbar1: Done

pybandits/base.py
```diff
@@ -91,10 +92,14 @@ class BaseMab(PyBanditsBaseModel, ABC):
         The list of possible actions, and their associated Model.
     strategy: Strategy
         The strategy used to select actions.
+    epsilon: Optional[Float01], defaults to 0
```
dariodandrea, line 95: Default is None, right?

pybandits/base.py
```diff
@@ -201,3 +206,47 @@ def get_state(self) -> (str, dict):
...
+        actions: Optional[Dict[ActionId, Model]] = None,
+    ) -> ActionId:
+        """
+        Wraps self.strategy.select_action function with epsilon-greedy strategy,
+        such that with probability epsilon a default_action is selected,
+        and with probability 1-epsilon the select_action function is triggered to choose action.
+        If no default_action is provided, a random action is selected.
+
+        Reference: Reinforcement Learning: An Introduction, Ch. 2 (Sutton and Burto, 2018)
+               https://web.stanford.edu/class/psych209/Readings/SuttonBartoIPRLBook2ndEd.pdf&ved=2ahUKEwjMy8WV9N2HAxVe0gIHHVjjG5sQFnoECEMQAQ&usg=AOvVaw3bKK-Y_1kf6XQVwR-UYrBY
```
dariodandrea, line 223: Minor comment: Please add 1 empty line after the reference link

pybandits/base.py
```diff
@@ -91,10 +92,14 @@ class BaseMab(PyBanditsBaseModel, ABC):
...
     strategy: Strategy
         The strategy used to select actions.
+    epsilon: Optional[Float01], defaults to 0
+        The probability of selecting a random action.
     """
 
     actions: Dict[ActionId, Model]
     strategy: Strategy
+    epsilon: Optional[Float01] = None
+    default_action: Optional[ActionId]
```
dariodandrea, line 102: in the code below (all the bandits classes) you set default_action to None. Is it what you want to do also here?

tests/test_smab.py
```diff
@@ -802,3 +829,82 @@ def test_smab_mo_cc_from_state(state):
...
+    st.integers(min_value=1),
+)
+def test_can_instantiate_epsilon_greddy_smab_with_params(a, b):
+    s = SmabBernoulli(
+        actions={
+            "action1": Beta(n_successes=a, n_failures=b),
+            "action2": Beta(n_successes=a, n_failures=b),
+        },
+        epsilon=0.1,
+        default_action="a1",
```
dariodandrea, line 851: here the initialization of this SmabBernoulli should raise an error: `"a1" is not in the actions set `

tests/test_smab.py
```diff
@@ -802,3 +829,82 @@ def test_smab_mo_cc_from_state(state):
...
+            "a5": Beta(),
+        },
+        epsilon=0.1,
+        default_action="a1",
+    )
+    forbidden_actions = set(["forb_1", "forb_2"])
+
+    _, _ = s.predict(n_samples=1000, forbidden_actions=forbidden_actions)
+
+
```
dariodandrea, line 874: If you can for all tests below try to avoid hardcoded values in the test body. But define them in the signature of the function (either with hypotesis or not)

pybandits/base.py
shaharbar1: Sure. So I'm also adding under the reference in MultiObjectiveBandit

pybandits/base.py
shaharbar1: In contrast to all child classes, this class is abstract, so there was no use to define default value. WDYT? In fact, I can remove the default from epsilon as well...

tests/test_smab.py
shaharbar1: Right! Gr8 catch!
I added relevant functionality under _get_valid_actions and a new root_validator for default_action under the BaseMAB.

tests/test_smab.py
shaharbar1: Please be more specific. If you refer to n_samples = 1000, I did it similarly to test_smab_mo_predict and other tests. Only difference is that I did not define a variable n_samples=1000 and passed it to the function

pybandits/base.py
dariodandrea: I agree. Let's remove the default in both as you did now.

tests/test_smab.py
dariodandrea: Just a minor comment: in `test_epsilon_greedy_smab_predict()` you do not define n_samples=1000 as variable, while in `test_epsilon_greddy_smab_mo_predict()` you define it in the first line of the function.
You can use the same approach (whatever you want) to keep all the code with the same style.

dariodandrea (approved):
Great job @shaharbar1 :)

PR #38, opened 2024-08-15
Add GitHub Actions workflow for release and tagging

shaharbar1:
Change log:
 1. Added bump_version.py to allow extracting version from toml.
 2. Added release.yml to allow github action of release and tag.

### DIFF
--- a/.github/workflows/continuous_delivery.yml
+++ b/.github/workflows/continuous_delivery.yml
@@ -1,0 +1,75 @@
+name: Publish and Tag
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  check_version:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ 3.8, 3.9 ]
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Poetry
+        run: |
+          curl -sSL https://install.python-poetry.org | python3 -
+          export PATH="$HOME/.poetry/bin:$PATH"
+
+      - name: Install project dependencies with Poetry
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
+name: Publish and Tag
```
dariodandrea, line 1: I would suggest to name this file continuous_delivery (so it's the other file is called continuous_integration)

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+
+on:
+  push:
+    branches:
+      - develop
```
dariodandrea, line 6: Let's create the branch 'main' and activate continuous_delivery when merged to 'main'

.github/workflows/release.yml
shaharbar1: Good suggestion. Done.

.github/workflows/release.yml
shaharbar1: I did this change on the yml.
Once the PR is accepted, I will open the branch, push develop into it, and tag it.

.github/workflows/continuous_delivery.yml
```diff
@@ -0,0 +1,58 @@
...
+          echo "PACKAGE_VERSION=${VERSION}" >> $GITHUB_ENV
+
+      - name: Verify if version has changed
+        id: version_check
+        run: |
+          if [ $(git tag -l "v${{ env.PACKAGE_VERSION }}") ]; then
+            echo "Version ${{ env.PACKAGE_VERSION }} already exists."
+            exit 0
+          fi
+
```
dariodandrea, line 39: Please add all the style checks and unit tests execution as done in continuous_integration.yml

shaharbar1: Done.
I also added "main" in CI config's PR triggers, since the CD only covers for changes in push.
I removed the dependency on changes in pyproject.toml, as this is already covered by the fi...end action under "Verify if version has changed"

dariodandrea (approved):
Good job @Shahar-Bar!

PR #61, opened 2024-09-24
Vectorization of cMAB Predict

shaharbar1:
### Changes
 * Changed predict method of BaseCmabBernoulli on cmab.py to vectorized version, rather than for loop.


dariodandrea (approved):
Very good! Just increment version in the toml file to 0.6.1

PR #65, opened 2024-10-14
Add loss factor to CostControlBandit and update logic for dynamic subsidy factor

shaharbar1:
### Changes:
* Added update of subsidy factor on MAB update step.
* Added option to allow Beta regime for subsidy factor and added convex combination loss function for determining best action over feasible set.
* Edited MultiObjectiveCostControlBandit on strategy.py to generalize CostControlBandit.
* Edited Base MAB, sMAB, and cMAB classes in mab.py, smab.py, and cmab.py, respectively to reduce duplicated code.
* Added test suite to support the new dynamic CC functionality.

### DIFF
--- a/pybandits/mab.py
+++ b/pybandits/mab.py
@@ -169,7 +169,6 @@ class BaseMab(PyBanditsBaseModel, ABC):
 
     ####################################################################################################################
 
-    @abstractmethod
     @validate_call
     def update(
         self, actions: List[ActionId], rewards: Union[List[BinaryReward], List[List[BinaryReward]]], *args, **kwargs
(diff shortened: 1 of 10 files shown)


tests/test_mab.py
```diff
@@ -39,14 +39,13 @@ class DummyMab(BaseMab):
     epsilon: Optional[Float01] = None
     default_action: Optional[ActionId] = None
 
-    def update(self, actions: List[ActionId], rewards: Union[List[BinaryReward], List[List[BinaryReward]]]):
-        self._validate_update_params(actions=actions, rewards=rewards)
+    def _update(self, actions: List[ActionId], rewards: Union[List[BinaryReward], List[List[BinaryReward]]]):
```
adarmiento, line 42: This type would benefit of being simplified in something like 
```
type BinaryRewards = List[BinaryReward]

def _update(...): List[BinaryReward] | List[BinaryRewards]

```

pybandits/mab.py
```diff
@@ -182,10 +181,27 @@ def update(
         rewards: List[Union[BinaryReward, List[BinaryReward]]]
             The reward for each sample.
         """
+        self._validate_update_params(actions=actions, rewards=rewards)
```
adarmiento, line 184: This side effect should be a function decorator

adarmiento (approved):

PR #69, opened 2024-12-11
Add epsilon greedy support in from_state

shaharbar1:
### Changes
 * Add epsilon and default_action to from_state methods in smab.py and cmab.py.


pybandits/cmab.py
dariodandrea, line 1: Don't we need to update from_state in cmab also for MO and MOCC?

shaharbar1: There is no cmab mo and mocc yet

PR #71, opened 2024-12-31
Adaptive Windowing for Multi-Armed Bandits

shaharbar1:
### Changes:
 * Added adaptive windowing mechanism to detect and handle concept drift in MAB modelsץ
 * Introduced ActionsManager class to handle action memory and updates with configurable window sizesץ
 * Refactored Model class hierarchy to support model resetting and memory managementץ
 * Added support for infinite and fixed-size windows with change detection via delta parameterץ
 * Enhanced test coverage for adaptive windowing functionality across MAB variantsץ

### DIFF
--- a/pybandits/actions_manager.py
+++ b/pybandits/actions_manager.py
@@ -1,0 +1,777 @@
+import warnings
+from abc import ABC, abstractmethod
+from collections import defaultdict, deque
+from inspect import isclass
+from typing import Any, Callable, Deque, Dict, Generic, List, Literal, Optional, Set, Union
+
+import numpy as np
+from numpy.typing import ArrayLike
+
+from pybandits.base import ACTION_IDS_PREFIX, ACTIONS, ActionId, BinaryReward, PositiveProbability, PyBanditsBaseModel
+from pybandits.model import (
+    BaseModel,
+    CmabModelType,
+    Model,
+    ModelMO,
+    SmabModelType,
+)
+from pybandits.pydantic_version_compatibility import (
+    PYDANTIC_VERSION_1,
+    PYDANTIC_VERSION_2,
+    GenericModel,
+    NonNegativeInt,
+    NonPositiveInt,
+    PositiveInt,
+    field_validator,
+    model_validator,
+    pydantic_version,
+    validate_call,
+)
+from pybandits.utils import extract_argument_names_from_function
+
+_NO_CHANGE_POINT = -1
+
+
+class ActionsManager(PyBanditsBaseModel, ABC):
+    """
+    Base class for managing actions and their associated models.
+    The class allows to account for non-stationarity by providing an adaptive window scheme for action update.
+    The change point detection is based on the adaptive windowing scheme.
+
+    Reference: Scaling Multi-Armed Bandit Algorithms (Fouché et al., 2019)
+               https://edouardfouche.com/publications/S-MAB_FOUCHE_KDD19.pdf
+
+    Parameters
+    ----------
+    actions : Dict[ActionId, Model]
+        The list of possible actions, and their associated Model.
+    adaptive_window_size : Optional[Union[PositiveInt, Literal["inf"]]]
+        The size of the adaptive window for action update. If None, no adaptive window is used.
+    delta : Optional[PositiveProbability], 0.1 if not specified.
+        The confidence level for the adaptive window.
+    """
+
+    actions: Dict[ActionId, BaseModel]
+    adaptive_window_size: Optional[Union[PositiveInt, Literal["inf"]]] = None
+    delta: Optional[PositiveProbability] = None
+
+    actions_memory: Optional[Deque] = None
+    rewards_memory: Optional[Deque] = None
+
+    if pydantic_version == PYDANTIC_VERSION_1:
+
+        class Config:
+            arbitrary_types_allowed = True
+            json_encoders = {deque: list}
+
+    elif pydantic_version == PYDANTIC_VERSION_2:
+        model_config = {"arbitrary_types_allowed": True, "json_encoders": {deque: list}}
+    else:
+        raise ValueError(f"Unsupported pydantic version: {pydantic_version}")
+
+    @field_validator("actions", mode="before")
+    @classmethod
+    def at_least_one_action_is_defined(cls, v):
+        # validate number of actions
+        if len(v) == 0:
+            raise AttributeError("At least one action should be defined.")
+        elif len(v) == 1:
+            warnings.warn("Only a single action was supplied. This MAB will be deterministic.")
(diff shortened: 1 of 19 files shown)


pybandits/actions_manager.py
```diff
@@ -0,0 +1,777 @@
...
+            Remaining number of elements from last memory state after current update.
+
+        Returns
+        -------
+        NonNegativeInt
+            The last change point for the given action. -1 if no change point is found.
+        """
+        action_index = np.where([a == action_id for a in self.actions_memory])[0].tolist()
+
+        rewards_window = [self.rewards_memory[i] for i in action_index]
```
j3rom3c, line 431: I remember reward encoding could be -1/1 ... is it an issue here

pybandits/actions_manager.py
```diff
@@ -0,0 +1,777 @@
...
+        window_length = len(rewards_window)
+        if window_length < 2:
+            return _NO_CHANGE_POINT
+        cumulative_reward = np.cumsum(np.array(rewards_window), axis=0)
```
j3rom3c, line 435: I may be wrong but reward may be encoded -1/1 instead of 0/1 (in Hive I think). Did you make sure that the encoding was correct for cumsum?

pybandits/actions_manager.py
```diff
@@ -0,0 +1,777 @@
...
+            The action ID.
+        residual_memory_len : Optional[NonNegativeInt]
+            Remaining number of elements from last memory state after current update.
+
+        Returns
+        -------
+        NonNegativeInt
+            The last change point for the given action. -1 if no change point is found.
+        """
+        action_index = np.where([a == action_id for a in self.actions_memory])[0].tolist()
```
j3rom3c, line 429: What is action_memory?

j3rom3c (commented):
I would need further details in the code if possible. Thx! (and sorry)

yana-gonnord (approved):
Amazing work!

pybandits/actions_manager.py
```diff
@@ -0,0 +1,777 @@
...
+
+        Returns
+        -------
+        Optional[PositiveInt]
+            The maximum length for the memory.
+        """
+        return self._get_max_len(self.adaptive_window_size)
+
+    @classmethod
+    def _get_expected_memory_length(cls, actions: Dict[ActionId, BaseModel]) -> NonNegativeInt:
```
Yoav-T, line 218: isn't this calculating the exact length of the memory? i.e. the number of treatments delivered? why do we call it "expected memory length" ?

pybandits/actions_manager.py
```diff
@@ -0,0 +1,777 @@
...
+            The reward for each sample.
+        """
+        self._validate_update_params(actions, rewards, **kwargs)
+
+        if self.adaptive_window_size is not None:
+            memory_len = len(self.actions_memory)
+            new_samples_len = len(actions)
+            if type(self.adaptive_window_size) is int:
+                residual_memory_len = max(
+                    0, memory_len - max(memory_len + new_samples_len - self.adaptive_window_size, 0)
```
Yoav-T, line 309: why do we need the internal max function? wouldn't the external 0 cover this case? that is, the case where the adaptive window size is infinity or a number larger than `memory_len + new_samples_len `

pybandits/actions_manager.py
Yoav-T: From what I can understand it's the history of actions taken within the adaptive window 

@shaharbar1 please correct me if I'm wrong

pybandits/actions_manager.py
Yoav-T: we can future-proof this with a cumsum over the boolean vector where it equals 1 e.g.:

`cumulative_reward = np.cumsum(np.array(rewards_window)==1, axis=0)`

that way it wouldn't matter

shaharbar1: > I may be wrong but reward may be encoded -1/1 instead of 0/1 (in Hive I think). Did you make sure that the encoding was correct for cumsum?

Yes.

shaharbar1: > we can future-proof this with a cumsum over the boolean vector where it equals 1 e.g.:
> 
> `cumulative_reward = np.cumsum(np.array(rewards_window)==1, axis=0)`
> 
> that way it wouldn't matter

No need to future proof. Rewards are validated via pydantic according to BinaryReward type.

pybandits/actions_manager.py
shaharbar1: residual_memory_len  = How many samples from past updates were left in the memory.

pybandits/actions_manager.py
shaharbar1: Based on the history of rewards, this counts how many treatments were given in the past. So this is the memory we would have expected to have. In practice our memory might be shorter.

pybandits/base.py
```diff
@@ -92,10 +96,43 @@ def _apply_version_adjusted_method(self, v2_method_name: str, v1_method_name: st
...
+            "model_copy", "copy", update={argument_name: argument_value}
+        )
+        return mutated_strategy
+
     @classmethod
     def _get_value_with_default(cls, key: str, values: Dict[str, Any]) -> Any:
         return values.get(key, cls.model_fields[key].default)
 
+    @classmethod
+    def _get_field_type(cls, key: str) -> Any:
```
Yoav-T, line 125: it's gonna be fun removing all this pydantic_version_1 stuff once we deprecate it

pybandits/mab.py
```diff
@@ -153,52 +153,46 @@ def _get_valid_actions(self, forbidden_actions: Optional[Set[ActionId]]) -> Set[
...
-        invalid = set(actions) - set(self.actions.keys())
-        if invalid:
-            raise AttributeError(f"The following invalid action(s) were specified: {invalid}.")
-        if len(actions) != len(rewards):
-            raise AttributeError(f"Shape mismatch: actions and rewards should have the same length {len(actions)}.")
+    def model_post_init(self, __context: Any) -> None:
+        if self.actions_manager.adaptive_window_size is not None and (
+            not self.epsilon or self.default_action is not None
+        ):
+            raise ValueError("Adaptive window size requires epsilon greedy super strategy with not default action.")
```
Yoav-T, line 173: why is it the case that adaptive window requires epsilon greedy with no default action? I couldn't figure it out

pybandits/base.py
shaharbar1: Damn right

Yoav-T (approved):
I can't say that I 100% understand the flow of the code, but I can't see anything wrong with the changes you've made. **Approved**

fsimond (approved):

PR #72, opened 2025-01-28
Add zooming quantitative bandit model

shaharbar1:
### Changes:
 * Added quantitative model support for continuous action spaces using zooming algorithm.
 * Added base model classes to separate single/multi-objective and cost control models.
 * Refactored MAB classes to support both discrete and continuous action spaces.
 * Updated test suite with new test cases for quantitative models and refactored test suite for robustness.
 * Added serialization support for quantitative models.

### DIFF
--- a/pybandits/mab.py
+++ b/pybandits/mab.py
@@ -19,7 +19,8 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-import json
+
+import random
 from abc import ABC, abstractmethod
 from inspect import isclass
 from typing import Any, Dict, List, Optional, Set, Union, get_origin
@@ -32,18 +33,22 @@ from pybandits.base import (
     ActionRewardLikelihood,
     BinaryReward,
     Float01,
+    MOProbability,
+    MOProbabilityWeight,
     PositiveProbability,
     Predictions,
+    Probability,
+    ProbabilityWeight,
     PyBanditsBaseModel,
+    Serializable,
+    UnifiedActionId,
 )
-from pybandits.model import BaseModel
+from pybandits.base_model import BaseModel
+from pybandits.model import Model, ModelMO
 from pybandits.pydantic_version_compatibility import (
-    PYDANTIC_VERSION_1,
-    PYDANTIC_VERSION_2,
-    model_validator,
-    pydantic_version,
     validate_call,
 )
+from pybandits.quantitative_model import QuantitativeModel
 from pybandits.strategy import Strategy
 from pybandits.utils import extract_argument_names_from_function
 
@@ -71,7 +76,7 @@ class BaseMab(PyBanditsBaseModel, ABC):
     actions_manager: ActionsManager
     strategy: Strategy
     epsilon: Optional[Float01] = None
-    default_action: Optional[ActionId] = None
+    default_action: Optional[UnifiedActionId] = None
 
     def __init__(
         self,
@@ -89,12 +94,18 @@ class BaseMab(PyBanditsBaseModel, ABC):
 
     @classmethod
     def _get_instantiated_class_attribute(cls, attribute_name: str, kwargs: Dict[str, Any]) -> PyBanditsBaseModel:
+        attribute_class = cls._get_attribute_type(attribute_name)
         if attribute_name in kwargs:
-            attribute = kwargs[attribute_name]
+            attribute = (
+                kwargs[attribute_name]
+                if isinstance(kwargs[attribute_name], attribute_class)
+                else attribute_class(**kwargs.pop(attribute_name))
+            )
         else:
-            attribute_class = cls._get_attribute_type(attribute_name)
-            required_sub_attributes = extract_argument_names_from_function(attribute_class.__init__, True)
-            if not required_sub_attributes:  # case of no native __init__ method, just pydantic generic __init__
+            required_sub_attributes = extract_argument_names_from_function(attribute_class.__init__)
+            if required_sub_attributes == extract_argument_names_from_function(
+                PyBanditsBaseModel.__init__
+            ):  # case of no native __init__ method, just pydantic generic __init__
                 required_sub_attributes = list(attribute_class.model_fields.keys())
                 sub_attributes = {k: kwargs.pop(k) for k in required_sub_attributes if k in kwargs}
             else:
@@ -108,31 +119,25 @@ class BaseMab(PyBanditsBaseModel, ABC):
 
     ############################################ Instance Input Validators #############################################
 
-    if pydantic_version == PYDANTIC_VERSION_1:
-
-        @model_validator(mode="before")
-        @classmethod
-        def check_default_action(cls, values):
-            epsilon = cls._get_value_with_default("epsilon", values)
-            default_action = cls._get_value_with_default("default_action", values)
-            if not epsilon and default_action:
-                raise AttributeError("A default action should only be defined when epsilon is defined.")
-            if default_action and default_action not in values["actions_manager"].actions:
-                raise AttributeError("The default action must be valid action defined in the actions set.")
-            return values
-
-    elif pydantic_version == PYDANTIC_VERSION_2:
-
-        @model_validator(mode="after")
-        def check_default_action(self):
-            if not self.epsilon and self.default_action:
-                raise AttributeError("A default action should only be defined when epsilon is defined.")
-            if self.default_action and self.default_action not in self.actions:
-                raise AttributeError("The default action must be valid action defined in the actions set.")
-            return self
-
-    else:
-        raise ValueError(f"Unsupported pydantic version: {pydantic_version}")
+    def model_post_init(self, __context: Any) -> None:
+        if self.actions_manager.delta is not None and (not self.epsilon or self.default_action is not None):
+            raise ValueError("Adaptive window requires epsilon greedy super strategy with not default action.")
+        if not self.epsilon and self.default_action:
+            raise AttributeError("A default action should only be defined when epsilon is defined.")
+        if self.default_action and self.default_action not in self.actions:
+            raise AttributeError("The default action must be valid action defined in the actions set.")
+        if (
+            self.default_action
+            and isinstance(self.default_action, tuple)
+            and not isinstance(self.actions[self.default_action[0]], QuantitativeModel)
+        ):
+            raise AttributeError("Quantitative default action requires a quantitative action model.")
+        if (
+            self.default_action
+            and isinstance(self.default_action, str)
+            and not isinstance(self.actions[self.default_action], (Model, ModelMO))
+        ):
+            raise AttributeError("Standard default action requires a standard action model.")
 
     ############################################# Method Input Validators ##############################################
 
@@ -165,10 +170,6 @@ class BaseMab(PyBanditsBaseModel, ABC):
 
     ####################################################################################################################
 
-    def model_post_init(self, __context: Any) -> None:
-        if self.actions_manager.delta is not None and (not self.epsilon or self.default_action is not None):
-            raise ValueError("Adaptive window requires epsilon greedy super strategy with not default action.")
-
     @property
     def actions(self) -> Dict[ActionId, BaseModel]:
         return self.actions_manager.actions
@@ -178,6 +179,7 @@ class BaseMab(PyBanditsBaseModel, ABC):
         self,
         actions: List[ActionId],
         rewards: Union[List[BinaryReward], List[List[BinaryReward]]],
+        quantities: Optional[List[Union[float, List[float], None]]] = None,
         actions_memory: Optional[List[ActionId]] = None,
         rewards_memory: Optional[Union[List[BinaryReward], List[List[BinaryReward]]]] = None,
         **kwargs,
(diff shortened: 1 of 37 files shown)


pybandits/model.py
```diff
@@ -80,30 +151,13 @@ class BaseBeta(Model):
         Counter of the number of failures.
     """
 
-    n_successes: PositiveInt = 1
-    n_failures: PositiveInt = 1
-
```
ronshiff1, line 96: Why did you take it to the parent class?

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,676 @@
...
+    """
+    Base class for quantitative models.
+
+    Parameters
+    ----------
+    dimension: PositiveInt
+        Number of parameters of the model.
+    """
+
+    dimension: PositiveInt
```
ronshiff1, line 34: why is the number of parameters of the model important?
(maybe add some description?)

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,676 @@
...
+    @abstractmethod
+    def sample_proba(self) -> float:
+        """
+        Sample the model.
+        """
+
+    @validate_call(config=dict(arbitrary_types_allowed=True))
+    def update(
+        self,
+        quantities: List[Union[float, List[float]]],
```
ronshiff1, line 45: isn't it should be:
Union[List[float], List[List[float]]]?
it can't be [3.5, 2.35, [1.3 , 23]] no? (this definition allows it)

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,676 @@
...
+        """
+
+        self._validate_params_lengths(quantities=quantities, rewards=rewards, **kwargs)
+        if quantities:
+            self._update(quantities, rewards, **kwargs)
+
+    @abstractmethod
+    def _update(
+        self,
+        quantities: Optional[List[Union[float, List[float], None]]],
```
ronshiff1, line 66: Same comment about the Order of List and Union

ronshiff1: another thing - can you please explain Union[float, List[float], None]

pybandits/model.py
```diff
@@ -482,22 +549,26 @@ def update(self, context: ArrayLike, rewards: List[BinaryReward]):
...
 
             # compute mean and std of the coefficients distributions
-            self.alpha.mu = mean(trace["alpha"])
-            self.alpha.sigma = std(trace["alpha"], ddof=1)
-            betas_mu = mean(trace["betas"], axis=0)
-            betas_std = std(trace["betas"], axis=0, ddof=1)
-            self.betas = [
-                StudentT(mu=mu, sigma=sigma, nu=beta.nu) for mu, sigma, beta in zip(betas_mu, betas_std, self.betas)
-            ]
+            if hasattr(trace, "alpha") and hasattr(trace, "betas"):
```
ronshiff1: why are there cases when there are no alpha or beta parameters in the trace?

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,676 @@
...
+        Parameters
+        ----------
+        quantities : Optional[List[Union[float, List[float], None]]
+            The value associated with each action. If none, the value is not used, i.e. non-quantitative action.
+        rewards: Union[List[BinaryReward], List[List[BinaryReward]]]
+            The reward for each sample.
+        """
+
+
+class QuantitativeModelCC(BaseModelCC, ABC):
```
ronshiff1, line 82: why does it have to inherit from ABC?

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,676 @@
...
+    cost: Callable[[Union[float, NonNegativeFloat]], NonNegativeFloat]
+        Cost associated to the Beta distribution.
+    """
+
+    cost: Callable[[Union[float, NonNegativeFloat]], NonNegativeFloat]
+
+
+class Segment(PyBanditsBaseModel):
+    """
+    Segment class.
```
ronshiff1: maybe it's worthwhile to elaborate a little bit?
maybe rename it to QuantitySegment? or something like that? 
Segment for me is a bit confusing with player segment (up to your decision of course)

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,676 @@
...
+
+    Parameters
+    ----------
+    intervals: Tuple[Tuple[Probability, Probability], ...]
+        Intervals of the segment.
+    """
+
+    intervals: Tuple[Tuple[Probability, Probability], ...]
```
ronshiff1: the intervals here are over the quantity space. 
even though it's [0,1] as Probability, I wouldn't use this typehint here... because it seems as it is interval over the probability space... 
and if maybe we decide so it's not [0,1] (something like [-0.5,0.5]. I would decouple and create another typehint

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,676 @@
...
+            Whether the segments are adjacent.
+        """
+        if self.intervals_array.shape[0] != other.intervals_array.shape[0]:
+            raise ValueError("Segments must have the same shape.")
+        return np.all(self.maxs == other.mins) or np.all(self.mins == other.maxs)
+
+
+class ZoomingModel(QuantitativeModel, ABC):
+    """
+    Base class for zooming models.
```
ronshiff1: Maybe a small explanation on the mechanism? 
what are the meanings of the parameters?

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,708 @@
...
+        segments: List[Segment],
+        rewards: List[BinaryReward],
+        **kwargs,
+    ):
+        segment_scores = {segment: model.mean for segment, model in self.segmented_actions.items()}
+        ordered_segments = sorted(segment_scores, key=segment_scores.get)
+        best_segment = ordered_segments[-1]
+        del self.sub_actions[best_segment.intervals]
+
+        # Consider merging adjacent segments
```
ronshiff1: I would consider splitting this and putting it in  methods with explanations

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,708 @@
...
+            Segment to check.
+
+        Returns
+        -------
+        bool
+            Whether the segments are adjacent.
+        """
+        if self.intervals_array.shape[0] != other.intervals_array.shape[0]:
+            raise ValueError("Segments must have the same shape.")
+        return np.all(self.maxs == other.mins) or np.all(self.mins == other.maxs)
```
ronshiff1: are you sure this logic is correct?
think of 2d intervals
then segments are adjust if in one dimension min = max and in the other dimension min=min and max=max, no?
image

pybandits/quantitative_model.py
ronshiff1: OK - I get it ... maybe call it quant_action_dim, or something like that?
I thought at the the beginning it's the cmab parameters...

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,762 @@
...
+            Rewards for update.
+        context : Optional[ArrayLike]
+            Context for update.
+        """
+
+    def _map_values_to_segments(self, quantities: List[Union[float, np.ndarray]]) -> List[Segment]:
+        segments = [segment for value in quantities for segment in self.segmented_actions.keys() if value in segment]
+        return segments
+
+    def _update_segmentation(
```
ronshiff1, line 403: docstring will be super helpful here.
The logic is not trivial

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,762 @@
...
+        quantities : List[Union[float, np.ndarray]]
+            The value associated with each action.
+        rewards: List[BinaryReward]
+            The reward for each sample.
+        context : Optional[ArrayLike]
+            Context for each sample.
+        """
+
+        segments = self._map_and_update_segment_models(quantities, rewards, **kwargs)
+        self._update_segmentation(quantities, segments, rewards, **kwargs)
```
j3rom3c, line 360: The _update_segmentation method is called without checking if segments is empty. Did you make sure that this could not happen?

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,762 @@
...
+        for segment, model in self.segmented_actions.items():
+            sampled_proba = model.sample_proba(**kwargs)
+            random_point = np.random.random((len(sampled_proba), len(segment.intervals)))
+            scaled_quantity = segment.mins.T + random_point * (segment.maxs.T - segment.mins.T)
+
+            result.append(tuple((tuple(quantity), prob) for quantity, prob in zip(scaled_quantity, sampled_proba)))
+        result = list(zip(*result))
+        return result
+
+    def _quantitative_update(self, quantities: List[Union[float, np.ndarray]], rewards: List[BinaryReward], **kwargs):
```
j3rom3c, line 345: The function adjust (split/merge) the segments and then update parameters using only last batch of data. Update (as I remember) was performed using information from a pickle file (in Alluxio) that contains all past aggregated data. For instance in sMAB the count of positive and negative rewards cumulated for a given action until now. You told us that Windowing uses only last batch of data, right? How/where do you access this limited info?

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,762 @@
...
+        -------
+        bool
+            Whether the segments have similar performance.
+        """
+        x = np.linspace(0, 1, self.n_comparison_points)
+        model1 = self.sub_actions[segment1.intervals]
+        model2 = self.sub_actions[segment2.intervals]
+        p1 = beta.pdf(x, model1.n_successes, model1.n_failures)
+        p2 = beta.pdf(x, model2.n_successes, model2.n_failures)
+        return jensenshannon(p1, p2) < self.comparison_threshold
```
j3rom3c, line 541: comparison_threshold = 0.1, by default. Did you have time to evaluate the default value and check its impact using simulation? Do you apply slit/merge window at the first update after a cold-start or is there a condition? I would expect a lot of merging during exploration phase.

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,762 @@
...
+            i += 1
+
+        # Split segments of interest if possible
+        i = 0
+        while i < len(interest_segments) - 1 and (
+            self.n_max_segments is None or len(self.sub_actions) < self.n_max_segments
+        ):
+            best_segment = interest_segments[i]
+            del self.sub_actions[best_segment.intervals]
+            sub_best_segments = best_segment.split()
```
j3rom3c, line 511: Would it be of interest to test new Beta distribution with Jensen-Shannon and validate split?

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,762 @@
...
+            **kwargs,
+        )
+
+    @classmethod
+    def _generate_initial_segments(cls, dimension: PositiveInt) -> List[Tuple[Tuple[Float01, Float01], ...],]:
+        interval_points = np.linspace(0, 1, cls._n_initial_segments + 1)
+        intervals = [(interval_points[i], interval_points[i + 1]) for i in range(cls._n_initial_segments)]
+        return [tuple(segment) for segment in product(intervals, repeat=dimension)]
+
+    def sample_proba(self, **kwargs) -> List[QuantitativeProbability]:
```
j3rom3c, line 331: I remember a discussion with Iurii explaining that not all values (in a segment) should be available to a user. For instance a user does not expect 1384647382 coins but may 1380000000 (some rounded value). Let's keep this in mine and check how we could round values in a segment and keep only "eligible" ones

j3rom3c (commented):
...

tests/test_quantitative_model.py
```diff
@@ -0,0 +1,227 @@
...
+    quantities = [boundary]
+    rewards = [1]
+    model.update(quantities=quantities, rewards=rewards)
+    mapped_segments = model._map_values_to_segments(quantities)
+    assert len(mapped_segments) >= 1
+
+
+# Test SmabZoomingModel initialization with valid parameters
+@given(dimension=st.integers(min_value=1, max_value=3))
+def initializes_smab_zooming_model_correctly(dimension):
```
ronshiff1: shouldn't it be def test_....?
and all the functions below as well

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,762 @@
...
+                    del self.sub_actions[segment.intervals]
+                    del self.sub_actions[other_segment.intervals]
+                    nuisance_segments.remove(segment)
+                    nuisance_segments.remove(other_segment)
+                    merged_segment = segment + other_segment
+                    self.sub_actions[merged_segment.intervals] = self._base_model.model_copy(deep=True)
+                    filtered_quantities, filtered_rewards, filtered_kwargs = self._filter_by_segment(
+                        merged_segment, quantities, segments, rewards, **kwargs
+                    )
+                    self._map_and_update_segment_models(filtered_quantities, filtered_rewards, **filtered_kwargs)
```
ronshiff1, line 478: in each such iteration you update the model.
what is the worst case number of updates?
O(k) where k is number of segments?

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,762 @@
...
+            sub_best_segments = best_segment.split()
+            self.sub_actions[sub_best_segments[0].intervals] = self._base_model.model_copy(deep=True)
+            self.sub_actions[sub_best_segments[1].intervals] = self._base_model.model_copy(deep=True)
+            filtered_quantities, filtered_rewards, filtered_kwargs = self._filter_by_segment(
+                best_segment, quantities, segments, rewards, **kwargs
+            )
+            self._map_and_update_segment_models(filtered_quantities, filtered_rewards, **filtered_kwargs)
+            i += 1
+
+    def is_similar_performance(self, segment1: Segment, segment2: Segment) -> bool:
```
ronshiff1, line 520: I'm not sure this is the right approach - 
let's say I have one small segment with small sample size (and large variance), and a segment with large sample size (and smaller variance) but they have similar mean - wouldn't we want to consider them as similar?
maybe hypothesis testing is more suitable here?
(calculate the probability to get the observations given that they have the underline distributions)

pybandits/mab.py
```diff
@@ -109,31 +119,28 @@ def _get_instantiated_class_attribute(cls, attribute_name: str, kwargs: Dict[str
...
-    elif pydantic_version == PYDANTIC_VERSION_2:
-
-        @model_validator(mode="after")
-        def check_default_action(self):
-            if not self.epsilon and self.default_action:
-                raise AttributeError("A default action should only be defined when epsilon is defined.")
-            if self.default_action and self.default_action not in self.actions:
-                raise AttributeError("The default action must be valid action defined in the actions set.")
-            return self
+    def model_post_init(self, __context: Any) -> None:
```
ronshiff1, line 122: I'm pretty sure that the model_post_init is ran twice-
once in PyBanditsBaseModel
and once in BaseModel.
maybe we can remove it from BaseModel as it inherits from PyBanditsBaseModel

pybandits/mab.py
```diff
@@ -270,15 +353,25 @@ def _select_epsilon_greedy_action(
...
+                if self.default_action:
+                    selected_action = self.default_action
+                else:
+                    selected_action = np.random.choice(
+                        np.array(list(set(a[0] if isinstance(a, tuple) else a for a in p.keys())), dtype=np.object_)
+                    )
+                    if isinstance(self.actions[selected_action], QuantitativeModel):
+                        selected_action = (
+                            selected_action,
+                            tuple(np.random.random(self.actions[selected_action].dimension)),
```
ronshiff1, line 383: nice!
something to consider - (not necessarily for this PR) - having the an arbitrary distribution (default can be uniform).

pybandits/mab.py
```diff
@@ -166,37 +173,114 @@ def _get_valid_actions(self, forbidden_actions: Optional[Set[ActionId]]) -> Set[
...
+            for key, value in action_probabilities.items()
+            if self._is_standard_action(value[0])
+        ]
+        list_transformations = self._transform_nested_list(list_transformations)
+        # Handle quantitative actions, for which the value is a tuple of
+        # tuples of (quantity, (probability, weight) or probability)
+        tuple_transformations = [
+            [{(key, quantity): proba for quantity, proba in sample} for sample in value]
+            for key, value in action_probabilities.items()
+            if self._is_quantitative_action(value[0])
```
ronshiff1, line 279: wouldn't it better that it will be a property of the model - (instead of inferring it)

pybandits/mab.py
```diff
@@ -166,37 +173,114 @@ def _get_valid_actions(self, forbidden_actions: Optional[Set[ActionId]]) -> Set[
...
+        if not list_transformations and not tuple_transformations:
+            return []
+        if not list_transformations:  # No standard actions
+            list_transformations = [dict() for _ in range(len(tuple_transformations))]
+        if not tuple_transformations:  # No quantitative actions
+            tuple_transformations = [dict() for _ in range(len(list_transformations))]
+        if not len(list_transformations) == len(tuple_transformations):
+            raise ValueError("The number of standard and quantitative actions should be the same.")
+        action_probabilities = [
+            {**list_dict, **dict_dict} for list_dict, dict_dict in zip(list_transformations, tuple_transformations)
```
ronshiff1: I'm completely confused by this - is it possible to give the variables informative names (what do they hold - e.g, quantitiative_actions_prob. something like that).

pybandits/strategy.py
```diff
@@ -225,31 +228,34 @@ def _average(cls, p_of_action: Union[Probability, List[Probability]]):
...
+        selected_action: UnifiedActionId
             The selected action.
         """
         # feasible actions enriched with their characteristics (cost, np.mean(probabilities), action_id)
-        # the negative probability ensures that if we order the actions based on their minimum values the one with
+        # the negative probability ensures that if we order the actions based on their minimum quantities the one with
         # higher probability will be selected
-        sortable_actions = [(actions[a].cost, -cls._average(p[a]), a) for a in feasible_actions]
+        sortable_actions = [
+            (actions[a[0]].cost(*a[1]) if isinstance(a, tuple) else actions[a].cost, -cls._average(p[a]), str(a))
```
ronshiff1: I feel we are relying to much on variable types in order to infer functionality....

ronshiff1: maybe it makes more sense to check whether cost is a constant or a callable

tests/test_cmab.py
ronshiff1, line 1: I can't say I understood all the changes - but basic question:
is all the tests that were removed are covered by the new tests?

pybandits/quantitative_model.py
```diff
@@ -0,0 +1,762 @@
...
+
+        Returns
+        -------
+        bool
+            Whether the segments have similar performance.
+        """
+        x = np.linspace(0, 1, self.n_comparison_points)
+        model1 = self.sub_actions[segment1.intervals]
+        model2 = self.sub_actions[segment2.intervals]
+        p1 = beta.pdf(x, model1.n_successes, model1.n_failures)
```
ronshiff1, line 451: I think that when working in CMAB this approach might be problematic - because it assumes that the distribution of the features in the segments is similar (and that's why counting successes is enough) - but if for example, we have two adjacent segments - for smaller quantity segment - there are users with a certain features distribution , and for the higher quantity segment there can be a different features distribution... 
even if they have the same beta distribution - I don't think it's correct to merge them

pybandits/model.py
```diff
@@ -674,13 +631,10 @@ class BayesianLogisticRegressionCC(BayesianLogisticRegression):
...
         chain Monte Carlo ("MCMC") or Variational Inference ("VI"). Check UpdateMethods in pybandits.model for the
         full list.
-    update_kwargs : Optional[dict], uses default values if not specified
+    update_kwargs : Optional[dict], uses default quantities if not specified
         Additional arguments to pass to the update method.
     cost: NonNegativeFloat
         Cost associated to the Bayesian Logistic Regression model.
     """
 
     cost: NonNegativeFloat
```
ronshiff1: is this required? isn't it inherited from ModelCC?

pybandits/quantitative_model.py
shaharbar1: Yes. It is verified

pybandits/quantitative_model.py
shaharbar1: Access to limited (last batch) data is only after split or merge for the new casted segments, e.g.  line #429.

pybandits/quantitative_model.py
shaharbar1: With simulation - it looked fine.
After first update, splits and merges won't happen, due to the other threshold that prevents segments from being marked as interesting or nuisance after first update.

pybandits/quantitative_model.py
shaharbar1: Possibly, but for alpha version of the method, I think we are cool.

pybandits/quantitative_model.py
shaharbar1: Definitely.

pybandits/quantitative_model.py
shaharbar1: I don't see why. It is an attribute of QuantitativeModel, so no connection with cMAB. The context should be clear.

pybandits/quantitative_model.py
shaharbar1: This is how it should be. For the same MAB, we could have one quantitative action with dim 1 and the other with dim 2.

pybandits/quantitative_model.py
shaharbar1: Explained above.

pybandits/model.py
shaharbar1: I encountered it in through the new test suite. Possibly if extreme values are passed as context, the update might fail. So, the trace would be empty.

pybandits/quantitative_model.py
shaharbar1: So you won't be able to instantiate it.

pybandits/quantitative_model.py
shaharbar1: Updating the docstring.

pybandits/quantitative_model.py
shaharbar1: I agree. I'll use Float01 instead.

tests/test_cmab.py
shaharbar1: That was the goal :-)

pybandits/model.py
shaharbar1: Indeed. removed it now.

pybandits/mab.py
shaharbar1: Done.

pybandits/model.py
shaharbar1: In order to keep track of number of successes and failures in all models.

pybandits/quantitative_model.py
shaharbar1: Sure.

pybandits/quantitative_model.py
shaharbar1: Nice catch! I'm changing np.all to np.any.

pybandits/quantitative_model.py
shaharbar1: Great point. Thanks!

tests/test_quantitative_model.py
shaharbar1: Indeed. Thanks

pybandits/quantitative_model.py
shaharbar1: I think so. It seems linear to me.

pybandits/mab.py
shaharbar1: @ronshiff1 It basically overrides the one in PyBanditsBaseModel. Why do you think it runs twice?

pybandits/mab.py
shaharbar1: @ronshiff1 But the model is not directly accessible at this point. We now have a list of recommendations that are not directly attached to a specific model. Of course I can get the model through the key and check its class but I don't see a major difference from the current approach.

pybandits/strategy.py
shaharbar1: @ronshiff1 I'm not seeing a big difference from the current approach - instead of relying on how the action looks, it will rely on how the cost looks.

ronshiff1 (approved):

PR #73, opened 2025-02-26
Feature/cmab bnn

ronshiff1:
- Added Bayesian Neural Network (BNN) with Fully connected layers as a an optional model for CMAB
- BayesianLogisiticRegression is now a sub-class of the BNN class (one layer BNN)
- LegacyBayesianLogisiticRegression still exists as reference (shoud be removed)
- Added a time comparison notebook - old and new BayesianLogisiticRegression sample_proba/update works at the same speed. BNN is slower and requires a more sophisticated optimizer and early stop (currently not included in the PR)
- Currently PR is not squashed.


notebooks/measure_run_time.ipynb
shaharbar1, line 1: Please remove this. I don't see a reason why this should be a part of the PR.

pybandits/cmab.py
```diff
@@ -54,21 +54,21 @@ class BaseCmabBernoulli(BaseMab):
...
 
-    actions: Dict[ActionId, BayesianLogisticRegression]
+    actions: Dict[ActionId, BayesianNeuralNetwork]
     predict_with_proba: bool
     predict_actions_randomly: bool
 
     @field_validator("actions", mode="after")
     @classmethod
-    def check_bayesian_logistic_regression_models(cls, v):
+    def check_bayesian_regression_models(cls, v):
```
shaharbar1: Make it check check_bayesian_neural_network_models

pybandits/cmab.py
```diff
@@ -43,7 +43,7 @@ class BaseCmabBernoulli(BaseMab):
 
     Parameters
     ----------
-    actions: Dict[ActionId, BayesianLogisticRegression]
+    actions: Dict[ActionId, BayesianNeuralNetwork]
```
shaharbar1, line 46: Make it BaseBayesianNeuralNetwork

pybandits/cmab.py
```diff
@@ -54,21 +54,21 @@ class BaseCmabBernoulli(BaseMab):
...
+    def check_bayesian_regression_models(cls, v):
         action_models = list(v.values())
         first_action = action_models[0]
         first_action_type = type(first_action)
         for action in action_models[1:]:
             if not isinstance(action, first_action_type):
                 raise AttributeError("All actions should follow the same type.")
-            if not len(action.betas) == len(first_action.betas):
-                raise AttributeError("All actions should have the same number of betas.")
+            if not action.expected_input == first_action.expected_input:
```
shaharbar1: Use a better name such as `input_dim`

pybandits/mab.py
```diff
@@ -311,6 +311,7 @@ def cold_start(
         action_ids: Optional[Set[ActionId]] = None,
         epsilon: Optional[Float01] = None,
         default_action: Optional[ActionId] = None,
+        action_model_class: Optional[Model] = None,
```
shaharbar1: Please explain the role of this parameter. Up until now this was redundant. Why do we need it?

pybandits/model.py
```diff
@@ -261,7 +263,64 @@ class StudentT(PyBanditsBaseModel):
     nu: confloat(allow_inf_nan=False) = 5.0
 
 
-class BayesianLogisticRegression(Model):
+class StudentTArray(PyBanditsBaseModel):
+    """
+    A model representing an array of Student's t-distributions.
```
shaharbar1: Please follow the docstring format as in all other files in the project.

shaharbar1: In addition the class attributes are hard to follow. Why do we have mu, sigma, and nu which seem to relate to a scalar student T inside an array of student T?
If the answer is to cover for scalar case, then these parameters seem redundant to me. Just make everything under the hood of params_dict.

shaharbar1: Also, correct me if I'm wrong, but `shape` feels like a cold start parameter rather then a class attribute.

pybandits/model.py
```diff
@@ -261,7 +263,64 @@ class StudentT(PyBanditsBaseModel):
...
+            nu = values["params_dict"].get("nu")
+
+            if not (mu and sigma and nu):
+                raise ValueError("params_dict must contain mu, sigma, and nu")
+            if not (np.array(mu).shape == np.array(sigma).shape == np.array(nu).shape):
+                raise ValueError("mu, sigma, and nu must have the same sizes")
+
+        return values
+
+    def __eq__(self, other):
```
shaharbar1: Specify return type. Specify other as Self.

pybandits/model.py
```diff
@@ -261,7 +263,64 @@ class StudentT(PyBanditsBaseModel):
...
+        for key, value in self.params_dict.items():
+            if not np.array_equal(value, other.params_dict[key]):
+                return False
+        return True
+
+    class Config:
+        arbitrary_types_allowed = True
+
+
+class LegacyBayesianLogisticRegression(Model):
```
shaharbar1: Please remove this.

pybandits/model.py
```diff
@@ -261,7 +263,64 @@ class StudentT(PyBanditsBaseModel):
...
+
+        return values
+
+    def __eq__(self, other):
+        for key, value in self.params_dict.items():
+            if not np.array_equal(value, other.params_dict[key]):
+                return False
+        return True
+
+    class Config:
```
shaharbar1: I don't think this is compatible with both v1 and v2 of pydantic

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
         )
 
 
-class BayesianLogisticRegressionCC(BayesianLogisticRegression):
+class BayesianNeuralNetwork(Model):
+    """Bayesian Neural Network model for binary classification.
```
shaharbar1, line 325: Again the docstring format.

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
+        Validate input context matrix dimensions
+    create_model(x, y, is_samplewise)
+        Create PyMC model for inference
+
+    Notes
+    -----
+    The network uses tanh activation for hidden layers and sigmoid for output layer.
+    Supported PyMC sampling methods are MCMC (default) and Variational Inference (VI).
+
+    Examples
```
shaharbar1: Add reference like in the legacy model for the paper that this class implements.

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
+        draws=1000,
+        chains=2,
+        init="adapt_diag",
+        cores=1,
+        target_accept=0.95,
+        progressbar=False,
+        return_inferencedata=False,
+    )
+
+    _default_variational_inference_fit_kwargs = dict(method="advi")
```
shaharbar1, line 368: I like this

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
+        def arrange_update_kwargs(cls, values):
+            update_kwargs = cls._get_value_with_default("update_kwargs", values)
+            update_method = cls._get_value_with_default("update_method", values)
+
+            if update_kwargs is None:
+                update_kwargs = dict()
+
+            if update_method == "VI":
+                update_kwargs["trace"] = {
+                    **cls._default_variational_inference_trace_kwargs,
```
shaharbar1, line 389: Create a utility function for this sort of dictionary merging, as it repeats in three different lines.

pybandits/model.py
```diff
@@ -261,7 +263,64 @@ class StudentT(PyBanditsBaseModel):
...
+
+            if not (mu and sigma and nu):
+                raise ValueError("params_dict must contain mu, sigma, and nu")
+            if not (np.array(mu).shape == np.array(sigma).shape == np.array(nu).shape):
+                raise ValueError("mu, sigma, and nu must have the same sizes")
+
+        return values
+
+    def __eq__(self, other):
+        for key, value in self.params_dict.items():
```
shaharbar1: all(np.array_equal(self.params_dict[key], other.params_dict[key]) for key in dict1)

pybandits/model.py
```diff
@@ -261,7 +263,64 @@ class StudentT(PyBanditsBaseModel):
...
+
+    """
+
+    shape: Optional[List[PositiveInt]] = None
+    params_dict: Optional[Dict[str, Union[List[float], List[List[float]]]]] = None
+    mu: confloat(allow_inf_nan=False) = 0.0
+    sigma: confloat(allow_inf_nan=False) = 10.0
+    nu: confloat(allow_inf_nan=False) = 5.0
+
+    @root_validator(pre=False, skip_on_failure=False)
```
shaharbar1: Also add validator for the keys of params_dict. On top of that, set in the type hint the key type to Literal rather than str.

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
+                self.update_kwargs["trace"] = {**self._default_mcmc_trace_kwargs, **self.update_kwargs.get("trace", {})}
+            else:
+                raise ValueError("Invalid update method.")
+            return self
+
+    else:
+        raise ValueError(f"Unsupported pydantic version: {pydantic_version}")
+
+    @classmethod
+    def create_posterior_params(cls, dim_list: List[PositiveInt]) -> List[Dict[str, StudentTArray]]:
```
shaharbar1: Again, output dict keys should be literal.

But in general the best practice IMO is to resort back to the original class of StudentT and return List[StudentT]

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
+        """
+        Returns the expected input dimension of the model.
+
+        Returns
+        -------
+        int
+            The number of input features expected by the model, derived from
+            the shape of the weight matrix in the first layer's posterior parameters.
+        """
+        return self.posterior_params[0]["w"].shape[0]
```
shaharbar1: Use a const for "w"

shaharbar1: and also for "b"

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
+            next_layer_input = bnn_input
+            for layer_ind in range(len(self.posterior_params)):
+                layer_params = self.posterior_params[layer_ind]
+                w_shape = np.array(layer_params["w"].params_dict["mu"]).shape  # without it n_features = 1 doesn't work
+                b_shape = np.array(layer_params["b"].params_dict["mu"]).shape
+
+                if is_sampelwise:
+                    w = PymcStudentT(
+                        f"w{layer_ind}", **layer_params["w"].params_dict, shape=(n_samples,) + w_shape
+                    )  # this create a tensor of shape (n_samples, n_features, n_output)
```
shaharbar1: Just write the shape in comment, no need for "this create a..."

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
+        Returns
+        -------
+        Tuple[Probability, float]
+            A tuple containing the sampled probabilities and the weighted sum.
+        """
+
+        # check input args
+        self.check_context_matrix(context=context)
+
+        _context = np.array(context, ndmin=2)
```
shaharbar1, line 594: why '_' in the name prefix?

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
+        dummy_y = np.zeros(len(context), dtype=np.int64)
+        _model = self.create_model(_context, dummy_y, is_sampelwise=True)
+
+        with _model:
+            trace = sample_prior_predictive(samples=1)
+
+        prob = trace["prior"]["prob"].values.reshape(-1)
+        weighted_sum = trace["prior"]["logit"].values.reshape(-1)
```
shaharbar1: Use a const for "logit" and "prob"

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
+            and initialization parameters
+        """
+
+        posterior_params = cls.create_posterior_params(dim_list)
+        return cls(
+            posterior_params=posterior_params, update_method=update_method, update_kwargs=update_kwargs, **kwargs
+        )
+
+    def __eq__(self, other):
+        for i, layer in enumerate(self.posterior_params):
```
shaharbar1: write it as loop over the zip of layers from both models.

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
-        Cost associated to the Bayesian Logistic Regression model.
+    if pydantic_version == PYDANTIC_VERSION_1:
+        betas: List[StudentT] = Field(..., min_items=1)
+    elif pydantic_version == PYDANTIC_VERSION_2:
+        betas: List[StudentT] = Field(..., min_length=1)
+    else:
+        raise ValueError("Invalid version.")
+
+    @model_validator(mode="before")
+    def set_posterior_params(cls, values):
```
shaharbar1: Why do we need extra functionality in the child class?

pybandits/model.py
```diff
@@ -536,30 +595,497 @@ def cold_start(
...
-        Student's t-distributions of the betas coefficients.
-    update_method : UpdateMethods, defaults to "MCMC"
-        The strategy for computing posterior quantities of the Bayesian models in the update function. Such as Markov
-        chain Monte Carlo ("MCMC") or Variational Inference ("VI"). Check UpdateMethods in pybandits.model for the
-        full list.
-    update_kwargs : Optional[dict], uses default values if not specified
-        Additional arguments to pass to the update method.
-    cost: NonNegativeFloat
-        Cost associated to the Bayesian Logistic Regression model.
+    if pydantic_version == PYDANTIC_VERSION_1:
```
shaharbar1: Why do we need betas? why isn't this just BNN with no hidden layers?

tests/test_cmab_simulator.py
```diff
@@ -54,7 +55,9 @@ def check_value_error(probs_reward, context):
 def test_cmab_e2e_simulation_with_default_arguments(
     action_ids=["a1", "a2"], n_features=3, n_updates=2, batch_size=10, num_groups=2
 ):
-    mab = CmabBernoulli.cold_start(action_ids=action_ids, n_features=n_features)
+    mab = CmabBernoulli.cold_start(
+        action_ids=action_ids, action_model_class=BayesianLogisticRegression, n_features=n_features
```
shaharbar1: Just use BNN all the way.

shaharbar1:
Please squeeze all your commits.

shaharbar1:
Please adapt your commit message to be similar to all the other commits.

shaharbar1 (changes requested):

pybandits/mab.py
ronshiff1: Until now, model class was taken from annotation - now - we have two possibilities - either Bayesian Logistic Regression or Bayesian Neural Network.

pybandits/model.py
ronshiff1: regarding the last two comments: 
shape, mu,sigma and nu are used for initialization - 
how would you do it? add a class method for init, and remove those class attributes?

pybandits/model.py
ronshiff1: why do we need to support the two versions?
if it's important we need to add a test for it

pybandits/model.py
ronshiff1: I don't know of a specific seminal paper.
What reference do you suggest:
1. https://arxiv.org/pdf/2007.06823
2. https://www.pymc.io/projects/examples/en/latest/variational_inference/bayesian_neural_network_advi.html

pybandits/model.py
ronshiff1: music to my ears

pybandits/model.py
ronshiff1: really? it's one line...(and it's relatively readable...)

pybandits/model.py
ronshiff1: done
if set(values["params_dict"].keys()) != {"mu", "sigma", "nu"}:
                raise ValueError("params_dict must contain mu, sigma, and nu")

pybandits/model.py
ronshiff1: what do you mean?

pybandits/model.py
ronshiff1: it's an internal variable inside the method. you have a better name?

pybandits/model.py
ronshiff1: again, I'm not sure what you mean with the const

pybandits/cmab.py
```diff
@@ -208,11 +208,19 @@ class CmabBernoulli(BaseCmabBernoulli):
         bandit strategy.
     """
 
-    actions: Dict[ActionId, BayesianLogisticRegression]
+    actions: Dict[ActionId, BayesianNeuralNetwork]
     strategy: ClassicBandit
     predict_with_proba: bool = False
     predict_actions_randomly: bool = False
 
+    def __eq__(self, other: "CmabBernoulli") -> bool:
```
adarmiento, line 216:
```suggestion
    def __eq__(self, other) -> bool:
          if type(self) is not type(other):
            return False
```

You can also return NotImplemented if you prefer, to let the interpreter know that equality is not defined and allow the other object's __eq__ method (if it exists) to handle the comparison.

pybandits/model.py
ronshiff1: This is to support initialization like:
BayesianLogisticRegression(alpha=StudentT(mu=..., ), betas=[StudentT(mu=..., ), StudentT(mu=..., ), StudentT(mu=..., )])

pybandits/model.py
ronshiff1: This is to have the same BLR as before. let's discuss

tests/test_cmab_simulator.py
ronshiff1: let's discuss

pybandits/cmab.py
ronshiff1: Thank you Alessandro for the feedback.

PR #77, opened 2025-03-16
Refactor adaptive windowing mechanism

shaharbar1:
### Changes:
 * Refactored adaptive windowing for multi-armed bandits to improve memory and state by removing memory from model internals. Instead, memory is supplied on the update method interface.
 * Updated test suite with new test cases for adaptive windowing and refactored test suite for robustness.

### DIFF
--- a/pybandits/actions_manager.py
+++ b/pybandits/actions_manager.py
@@ -2,7 +2,7 @@ import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict, deque
 from inspect import isclass
-from typing import Any, Callable, Deque, Dict, Generic, List, Literal, Optional, Set, Union
+from typing import Any, Callable, ClassVar, Dict, Generic, List, Optional, Set, Tuple, Union
 
 import numpy as np
 from numpy.typing import ArrayLike
@@ -21,15 +21,11 @@ from pybandits.pydantic_version_compatibility import (
     GenericModel,
     NonNegativeInt,
     NonPositiveInt,
-    PositiveInt,
     field_validator,
-    model_validator,
     pydantic_version,
     validate_call,
 )
 from pybandits.utils import extract_argument_names_from_function
-
-_NO_CHANGE_POINT = -1
 
 
 class ActionsManager(PyBanditsBaseModel, ABC):
@@ -45,18 +41,15 @@ class ActionsManager(PyBanditsBaseModel, ABC):
     ----------
     actions : Dict[ActionId, Model]
         The list of possible actions, and their associated Model.
-    adaptive_window_size : Optional[Union[PositiveInt, Literal["inf"]]]
-        The size of the adaptive window for action update. If None, no adaptive window is used.
-    delta : Optional[PositiveProbability], 0.1 if not specified.
-        The confidence level for the adaptive window.
+    delta : Optional[PositiveProbability]
+        The confidence level for the adaptive window. None for skipping the change point detection.
     """
 
     actions: Dict[ActionId, BaseModel]
-    adaptive_window_size: Optional[Union[PositiveInt, Literal["inf"]]] = None
     delta: Optional[PositiveProbability] = None
-
-    actions_memory: Optional[Deque] = None
-    rewards_memory: Optional[Deque] = None
+    _no_change_point: ClassVar[NonNegativeInt] = -1
+    _min_adaptive_window_size: ClassVar[NonPositiveInt] = 10000
+    _memory_parameters_suffix: ClassVar[str] = "_memory"
 
     if pydantic_version == PYDANTIC_VERSION_1:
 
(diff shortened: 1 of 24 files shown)


pybandits/actions_manager.py
```diff
@@ -367,24 +250,39 @@ def _update_actions(
         """
         pass
 
-    def _get_last_change_point(self, residual_memory_len: Optional[NonNegativeInt]) -> NonNegativeInt:
+    def _get_last_change_point(
```
ronshiff1, line 390: why don't we allow reset of one action?

ronshiff1:
A general question regarding the mechanism:
so we have all the past reward data in a table outside pybandits - 
how do you see the interface between the outside system (e.g, optx) and pybandits?
Does it sends all the past information in the update? without taking into account past detected change points?

pybandits/actions_manager.py
```diff
@@ -298,62 +179,64 @@ def update(self, actions: List[ActionId], rewards: Union[List[BinaryReward], Lis
...
-                )
-                self.rewards_memory = deque(
-                    (self.rewards_memory[i] for i in range(relative_change_point, 0)),
-                    maxlen=self.max_len,
-                )
+                last_change_point := self._get_last_change_point(residual_memory_len, actions_memory, rewards_memory)
+            ) != self._no_change_point:
+                relative_change_point = last_change_point - len(actions_memory)
+                actions_memory = actions_memory[relative_change_point:]
+                rewards_memory = rewards_memory[relative_change_point:]
```
ronshiff1: we are searching change points only in the memory - but can't it occur inside the "current" context?
it will be recognized only in this next update, no?

pybandits/actions_manager.py
```diff
@@ -45,18 +41,15 @@ class ActionsManager(PyBanditsBaseModel, ABC):
...
     """
 
     actions: Dict[ActionId, BaseModel]
-    adaptive_window_size: Optional[Union[PositiveInt, Literal["inf"]]] = None
     delta: Optional[PositiveProbability] = None
-
-    actions_memory: Optional[Deque] = None
-    rewards_memory: Optional[Deque] = None
+    _no_change_point: ClassVar[NonNegativeInt] = -1
+    _min_adaptive_window_size: ClassVar[NonPositiveInt] = 10000
```
ronshiff1, line 51: don't we need to define also a maximal size for the window? for bounding the complexity?

pybandits/actions_manager.py
```diff
@@ -418,50 +320,46 @@ def _get_last_change_point_for_action(
...
         -------
         NonNegativeInt
             The last change point for the given action. -1 if no change point is found.
         """
-        action_index = np.where([a == action_id for a in self.actions_memory])[0].tolist()
+        action_index = np.nonzero([a == action_id for a in actions_memory])[0].tolist()
 
-        rewards_window = [self.rewards_memory[i] for i in action_index]
+        rewards_window = [rewards_memory[i] for i in action_index]
         window_length = len(rewards_window)
```
ronshiff1, line 473: I think that it can be implemented more efficiently using numpy (instead of looping)
rewards_memory[actions_memory == a]

pybandits/actions_manager.py
```diff
@@ -409,7 +307,11 @@ def _get_threshold(self, past_trials: np.ndarray, present_trials: np.ndarray) ->
         return threshold
 
     def _get_last_change_point_for_action(
```
ronshiff1, line 446: I believe that numba can be beneficial here (since it has looping)

ronshiff1 (changes requested):

pybandits/actions_manager.py
shaharbar1: This is based on the AdWin paper. The idea is that given a change in one action, we prefer to reset all actions to assure stationary data.

pybandits/actions_manager.py
shaharbar1: It can. Do note that on line #204 we are extending the memory with current context.

pybandits/actions_manager.py
shaharbar1: I don't think so, because it is a matter of platform. The end user (MLE) should test that.

shaharbar1:
> A general question regarding the mechanism: so we have all the past reward data in a table outside pybandits - how do you see the interface between the outside system (e.g, optx) and pybandits? Does it sends all the past information in the update? without taking into account past detected change points?

@ronshiff1 very good point. I added memory trimming based on expected memory length.

pybandits/actions_manager.py
shaharbar1: I think you are right in general but not in our production use case. In our use case, the MAB runs update once in a while and the update starts from the state, not from an object that is kept in the memory. So JIT compilation will take its time, with no actuial added value for few actions.

pybandits/actions_manager.py
shaharbar1: I didn't test, but moving to numpy requires new memory allocation, so depending on the size it might not be beneficial.

pybandits/actions_manager.py
ronshiff1: They say that it can be local... but whatever you decide
image

ronshiff1 (approved):

PR #89, opened 2025-04-22
Add bayesian neural network model

ronshiff1:
### Changes

- Created a bayesian neural network (BNN) model for CMAB.
- The bayesian logistic regression implementation is based on the BNN class.
- Added a BNN tutorial notebook.
- Removed python 3.8 support.
- Removed deprecated notebooks.
- Updated pre commit hook to remove kernels from notebooks metadata.


pybandits/cmab.py
```diff
@@ -45,7 +45,7 @@ class BaseCmabBernoulli(BaseMab, ABC):
 
     Parameters
     ----------
-    actions: Dict[ActionId, BayesianLogisticRegression]
+    actions: Dict[ActionId, BayesianNeuralNetwork]
```
shaharbar1: Change to 'BaseBayesianNeuralNetwork'

pybandits/cmab.py
```diff
@@ -206,6 +206,16 @@ class CmabBernoulli(BaseCmabBernoulli):
     predict_with_proba: bool = False
     predict_actions_randomly: bool = False
 
+    def __eq__(self, other: "CmabBernoulli") -> bool:
```
shaharbar1: Are you sure this is needed? Isn't it native with pydantic?

pybandits/model.py
```diff
@@ -352,60 +354,155 @@ class BetaMOCC(BetaMO):
...
+    """
+
+    mu: Union[List[float], List[List[float]]]
+    sigma: Union[List[NonNegativeFloat], List[List[NonNegativeFloat]]]
+    nu: Union[List[PositiveFloat], List[List[PositiveFloat]]]
+
+    @model_validator(mode="after")
+    @classmethod
+    def validate_inputs(cls, values):
+        if pydantic_version == PYDANTIC_VERSION_1:
```
shaharbar1, line 405: Why do you need .get for pydantic 1?

pybandits/model.py
```diff
@@ -352,60 +354,155 @@ class BetaMOCC(BetaMO):
 SmabModelType = TypeVar("SmabModelType", bound=Union[BaseBeta, BetaMO])
 
 
-class StudentT(PyBanditsBaseModel):
+class StudentTArray(PyBanditsBaseModel):
     """
-    Student's t-distribution.
+    A class representing an array of Student's t-distributions with parameters `mu`, `sigma`, and `nu`.
```
shaharbar1, line 362: 1. Add empty line before "Attributes".
2. Please elaborate on what does 1d means vs. 2d.
3. Like I mentioned in our oral talk, I think you should make a class for as single layer (1d case?) separately from the entire BNN (2d case?). The 2d case should contain a list of 1d classes.

pybandits/model.py
```diff
@@ -352,60 +354,155 @@ class BetaMOCC(BetaMO):
...
+
+    @model_validator(mode="after")
+    @classmethod
+    def validate_inputs(cls, values):
+        if pydantic_version == PYDANTIC_VERSION_1:
+            mu_arr = np.array(values.get("mu"))
+            sigma_arr = np.array(values.get("sigma"))
+            nu_arr = np.array(values.get("nu"))
+        elif pydantic_version == PYDANTIC_VERSION_2:
+            mu_arr = np.array(values.mu)
```
shaharbar1: Here you make assumption that in the 2d case, all inner rows are of the same length. You should validate it.

pybandits/model.py
```diff
@@ -352,60 +354,155 @@ class BetaMOCC(BetaMO):
...
+
+    @property
+    def shape(self) -> Tuple[PositiveInt]:
+        return np.array(self.mu).shape
+
+    @property
+    def params(self):
+        return dict(mu=np.array(self.mu), sigma=np.array(self.sigma), nu=np.array(self.nu))
 
+    def __eq__(self, other: Self) -> bool:
```
shaharbar1: Again, are you sure this is needed, isn't it native in pydantic?

pybandits/model.py
```diff
@@ -352,60 +354,155 @@ class BetaMOCC(BetaMO):
...
+        return self.mu == other.mu and self.sigma == other.sigma and self.nu == other.nu
+
+
+class BnnLayerParams(PyBanditsBaseModel):
```
shaharbar1: Add docstring

pybandits/model.py
```diff
@@ -352,60 +354,155 @@ class BetaMOCC(BetaMO):
...
     """
 
-    mu: confloat(allow_inf_nan=False) = 0.0
-    sigma: confloat(allow_inf_nan=False) = 10.0
-    nu: confloat(allow_inf_nan=False) = 5.0
+    bnn_layer_params: List[BnnLayerParams]
+    bnn_layer_params_init: List[BnnLayerParams]
 
+    @staticmethod
+    def compare_layer_params(layer_params_self: List[BnnLayerParams], layer_params_other: List[BnnLayerParams]) -> bool:
```
shaharbar1: Add docstring

pybandits/model.py
shaharbar1: Attributes should become Parameters

pybandits/model.py
```diff
@@ -352,60 +354,155 @@ class BetaMOCC(BetaMO):
...
+        if not self.compare_layer_params(self.bnn_layer_params_init, other.bnn_layer_params_init):
+            return False
 
-    It is modeled as:
+        return True
 
-        y = sigmoid(alpha + beta1 * x1 + beta2 * x2 + ... + betaN * xN)
 
-    where the alpha and betas coefficients are Student's t-distributions.
+class BaseBayesianNeuralNetwork(Model, ABC):
```
shaharbar1, line 503: What about references?

pybandits/model.py
```diff
@@ -454,26 +575,48 @@ def arrange_update_kwargs(self):
...
-        x : Union[np.ndarray, TensorVariable]
-            Input values.
-
+        n_features : PositiveInt
+            The number of input features for the BNN.
+        hidden_dim_list : List[PositiveInt]
+            A list of integers specifying the number of hidden units in each hidden layer.
+            If None, no hidden layers are added.
+        **dist_params_init : dict, optional
+            Additional parameters for initializing the distribution of weights and biases.
```
shaharbar1, line 640: Missing empty line

pybandits/model.py
```diff
@@ -490,213 +633,288 @@ def check_context_matrix(self, context: np.ndarray):
...
-            Weighted sums between contextual feature values and sampled coefficients.
+        PymcModel
+            PyMC model object with the specified neural network architecture
+
+        Notes
+        -----
+        The model structure follows these steps:
+        1. For each layer, create weight and bias variables from StudentT distributions
+        2. Apply linear transformations and activations through the layers.
+           When is_sampelwise is True, the linear transformation is applied on each row separately (so random variables are not shared)
```
shaharbar1: Missing '.' at the end of the sentence.

shaharbar1: When will we use is_sampelwise =True? when will we use is_sampelwise =False? please add to the docstring.

pybandits/model.py
```diff
@@ -490,213 +633,288 @@ def check_context_matrix(self, context: np.ndarray):
...
-            # compute mean and std of the coefficients distributions
-            self.alpha.mu = mean(trace["alpha"])
-            self.alpha.sigma = std(trace["alpha"], ddof=1)
-            betas_mu = mean(trace["betas"], axis=1)
-            betas_std = std(trace["betas"], axis=1, ddof=1)
-            self.betas = [
-                StudentT(mu=mu, sigma=sigma, nu=beta.nu) for mu, sigma, beta in zip(betas_mu, betas_std, self.betas)
-            ]
+        for layer_ind in range(len(self.model_params.bnn_layer_params)):
+            name = f"weight_{layer_ind}"
```
shaharbar1: use a const / classvar instead of "weight" and "bias"

pybandits/model.py
```diff
@@ -490,213 +633,288 @@ def check_context_matrix(self, context: np.ndarray):
...
-            chain Monte Carlo ("MCMC") or Variational Inference ("VI"). Check UpdateMethods in pybandits.model for the
-            full list.
-        update_kwargs : Optional[dict], uses default values if not specified
-            Additional arguments to pass to the update method.
-        kwargs: Dict[str, Any]
-            Additional arguments for the Bayesian Logistic Regression child model.
+            Number of input features for the network.
+        hidden_dim_list : Optional[List[PositiveInt]], optional
+            List of dimensions for the hidden layers of the network. If None, no hidden layers are added.
+        update_method : UpdateMethods, optional
```
shaharbar1: Is it really optional?

pybandits/model.py
```diff
@@ -490,213 +633,288 @@ def check_context_matrix(self, context: np.ndarray):
...
+        Reset the model.
+        """
+        self.model_params.bnn_layer_params = deepcopy(self.model_params.bnn_layer_params_init)
 
+    def __eq__(self, other: Self) -> bool:
+        return self.model_params == other.model_params
 
-class BayesianLogisticRegression(BaseBayesianLogisticRegression):
+
+class BaseBayesianNeuralNetworkCC(BaseBayesianNeuralNetwork, ABC):
```
shaharbar1: Why do you need this base class?
Can't BayesianNeuralNetworkCC inherit from BaseBayesianNeuralNetwork and ModelCC?

pybandits/model.py
```diff
@@ -490,213 +633,288 @@ def check_context_matrix(self, context: np.ndarray):
...
+class BaseBayesianLogisticRegression(BaseBayesianNeuralNetwork, ABC):
+    """
+    A Bayesian Logistic Regression model that inherits from BaseBayesianNeuralNetwork.
+    This model is a specialized version of a Bayesian Neural Network with a single layer,
+    designed specifically for logistic regression tasks. The model parameters are
+    validated to ensure that the model adheres to this single-layer constraint.
+    """
 
-    where the alpha and betas coefficients are Student's t-distributions.
+    @field_validator("model_params")  # will be enabled in pydantic 2
```
shaharbar1: Remove this comment about pydantic

pybandits/model.py
```diff
@@ -352,60 +354,155 @@ class BetaMOCC(BetaMO):
...
     ----------
-    mu: float
-        Mean of the Student's t-distribution.
-    sigma: float
-        Standard deviation of the Student's t-distribution.
-    nu: float
-        Degrees of freedom.
+    bnn_layer_params : List[BnnLayerParams]
+        A list of BNN layer parameters representing the current state of the model.
+    bnn_layer_params_init : List[BnnLayerParams]
```
shaharbar1, line 480: Why do we need to keep the initialization params all the way?

tests/test_utils.py
```diff
@@ -24,9 +25,22 @@ def to_temporary_pickle(model: PyBanditsBaseModel):
...
 
     def sample(self, *args, **kwargs) -> Dict[str, np.ndarray]:
-        return {
-            "alpha": np.random.random(size=(1, self.n_draws)),
-            "betas": np.random.random(size=(self.n_features, self.n_draws)),
-        }
+        sample_dict = {}
+
+        dim_list = [self.n_features] + self.hidden_dim_list + [1]
+        for i in range(len(dim_list) - 1):
```
shaharbar1, line 35: Why do you need to mock the intermediate BNN outputs?

shaharbar1 (changes requested):

pybandits/model.py
ronshiff1: you are right. I removed it

pybandits/cmab.py
ronshiff1: you are right. removed

pybandits/model.py
ronshiff1: for pydantic 1 "values" is a dict. 
I could also use values["mu"], which is the same.
do you want me to change it?

pybandits/model.py
ronshiff1: 1. Fixed.
2. 1D is a list, and 2D is a list of lists... which is obvious from the typing... I'm not sure how can I elaborate...
3. I think that my approach is better. If it's a show stopper, let's discuss again. if not, let's keep it that way

pybandits/model.py
ronshiff1: done

pybandits/model.py
ronshiff1: done

pybandits/model.py
ronshiff1: there isn't really a reference to what we did I think

pybandits/model.py
ronshiff1: done

pybandits/model.py
ronshiff1: done

pybandits/model.py
ronshiff1: fixed

pybandits/model.py
ronshiff1: I think that ModelCC was introduced in your quantitive pr. no?
I didn't find it in the develop

pybandits/model.py
ronshiff1: done

pybandits/model.py
ronshiff1: this is for reseting in adwin.
in my implementation of cold start you can define the distributions intialization. so we need to track that if we want to reset

tests/test_utils.py
ronshiff1: they are being read in the the update. like here:

 for layer_ind in range(len(self.model_params.bnn_layer_params)):
      name = f"weight_{layer_ind}"
      w_mu = np.mean(trace[name], axis=0)
      w_sigma = np.std(trace[name], axis=0)

so we need them of course.

pybandits/model.py
ronshiff1: done

pybandits/model.py
ronshiff1: done

docs/tutorials/bnn_example.ipynb
```diff
@@ -0,0 +1,377 @@
...
+    "grid_label = np.ones(grid_2d.shape[0], dtype=np.int8).tolist()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "_model = blr.create_model(grid_2d, grid_label, is_samplewise=True)\n",
```
shaharbar1: Please add some markdown cells to explain your steps, e.g. Now create a model and than generate predictions.

docs/tutorials/cmab.ipynb
```diff
@@ -161,19 +166,17 @@
...
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {
   "hide_input": false,
+  "kernelspec": {
```
shaharbar1: Please remove this  kernelspec

pybandits/model.py
shaharbar1: 2. I understand the practice of 1d vs. 2d. I mean - when wouls one use 2d? when would one use 1d?
3. It's not a show stopper, but I just want the code right and self explanatory. Let's talk about this again.

pybandits/model.py
shaharbar1: Yes, that way you won't need the "if...else" statement.

pybandits/model.py
shaharbar1: OK, so please include that in the docstring...

shaharbar1: Also, make the field frozen so no one can set its value.

pybandits/model.py
shaharbar1: Maybe Neal, Radford (1996) – Bayesian Learning for Neural Networks?

pybandits/model.py
```diff
@@ -352,60 +355,213 @@ class BetaMOCC(BetaMO):
...
+    _prob_var_name: ClassVar[str] = "prob"
+    _weight_var_name: ClassVar[str] = "weight"
+    _bias_var_name: ClassVar[str] = "bias"
+
+    update_method: str = "MCMC"
     update_kwargs: Optional[dict] = None
-    _default_update_kwargs = dict(draws=1000, progressbar=False, return_inferencedata=False)
-    _default_mcmc_kwargs = dict(
+
+    _default_mcmc_trace_kwargs = dict(
```
shaharbar1: Add ClassVar type hint

pybandits/model.py
```diff
@@ -490,213 +690,311 @@ def check_context_matrix(self, context: np.ndarray):
...
         """
 
         # check input args
         self.check_context_matrix(context=context)
 
-        # extend context with a column of 1 to handle the dot product with the intercept
-        context_ext = c_[ones((len(context), 1)), context]
+        _context = np.array(context, ndmin=2)
+        dummy_y = np.zeros(len(context), dtype=np.int64)
+        _model = self.create_model(_context, dummy_y, is_samplewise=True)
```
shaharbar1: Instead of the dummy, could you make the 2nd parameter Optional, such that None makes it create the dummy value internally?

pybandits/model.py
```diff
@@ -454,26 +634,46 @@ def arrange_update_kwargs(self):
         raise ValueError(f"Unsupported pydantic version: {pydantic_version}")
 
     @classmethod
-    def _stable_sigmoid(cls, x: Union[np.ndarray, TensorVariable]) -> Union[np.ndarray, TensorVariable]:
```
shaharbar1, line 457: Don't we need the stable sigmoid anymore? You are still using sigmoid for the BNN output.

pybandits/model.py
```diff
@@ -490,213 +690,311 @@ def check_context_matrix(self, context: np.ndarray):
...
+                for dim, size in enumerate(x_expanded.shape):
+                    if size == 1:
+                        bnn_input = specify_broadcastable(bnn_input, dim)
+            else:
+                bnn_input = MutableData("bnn_input", x)
+
+            n_samples = len(x)
+
+            next_layer_input = bnn_input
+            for layer_ind in range(len(self.model_params.bnn_layer_params)):
```
shaharbar1: Should you just enumerate self.model_params.bnn_layer_params?

pybandits/model.py
```diff
@@ -490,213 +690,311 @@ def check_context_matrix(self, context: np.ndarray):
...
         ----------
-        context : np.ndarray
-            Context matrix of shape (n_samples, n_features).
+        x : ArrayLike
+            Input features of shape (n_samples, n_features)
+        y : Union[List[BinaryReward], np.ndarray]
+            Binary target values of shape (n_samples,)
+        is_sampelwise : bool
+            If True, process samples independently. If False, process all samples at once.
+            In the predict step, we would like to sample the model parameters independently for each sample (hench is_sampelwise=True).
```
shaharbar1: On the predict step case, your output is also Deterministic, maybe you can elaborate further on the difference, and maybe give the variable a more meaningful name - is_update? is_random? is_samplewise is a little too simplistic.

pybandits/model.py
```diff
@@ -490,213 +690,311 @@ def check_context_matrix(self, context: np.ndarray):
...
-            # Likelihood (sampling distribution) of observations
-            weighted_sum = Deterministic("weighted_sum", alpha + dot(betas, context.T))
-            p = Deterministic("p", self._stable_sigmoid(weighted_sum))
-
-            # Bernoulli random vector with probability of success given by sigmoid function and actual data as observed
-            _ = Bernoulli("likelihood", p=p, observed=rewards)
+        if len(context) != len(rewards):
+            raise AttributeError("Shape mismatch: context and rewards must have the same length.")
 
+        _context = np.array(context, ndmin=2)
```
shaharbar1: context is already coming as np.array. Use https://numpy.org/doc/stable/reference/generated/numpy.atleast_2d.html for the dimensions.

pybandits/model.py
```diff
@@ -490,213 +690,311 @@ def check_context_matrix(self, context: np.ndarray):
...
+        _model = self.create_model(x=_context, y=rewards, is_samplewise=False)
+        with _model:
             # update traces object by sampling from posterior distribution
             if self.update_method == "VI":
                 # variational inference
                 update_kwargs = self.update_kwargs.copy()
-                approx = fit(method=update_kwargs.pop("method"))
-                trace = approx.sample(**update_kwargs)
+                self._approx = fit(**update_kwargs["fit"])
```
shaharbar1, line 849: It looks like you are only using self._approx in the notebook example. If this is the case, dismiss it.

pybandits/model.py
shaharbar1: OK, but still BayesianNeuralNetworkCC should just inherit from BaseBayesianNeuralNetwork

pybandits/model.py
```diff
@@ -490,213 +690,311 @@ def check_context_matrix(self, context: np.ndarray):
...
+    This class implements a Bayesian Neural Network with cost control by extending
+    the BaseBayesianNeuralNetworkCC. It provides functionality for probabilistic
+    modeling and inference using neural networks with cost constraints.
     """
-    Bayesian Logistic Regression model with cost control.
 
-    It is modeled as:
 
-        y = sigmoid(alpha + beta1 * x1 + beta2 * x2 + ... + betaN * xN)
+class BaseBayesianLogisticRegression(BaseBayesianNeuralNetwork, ABC):
```
shaharbar1: No need for this base class. BayesianLogisticRegression should inherit BayesianNeuralNetwork, and similarly for the CCs.

pybandits/model.py
```diff
@@ -490,213 +690,311 @@ def check_context_matrix(self, context: np.ndarray):
...
+            w_mu = np.mean(trace[name], axis=0)
+            w_sigma = np.std(trace[name], axis=0)
+            self.model_params.bnn_layer_params[layer_ind].weight = StudentTArray(
+                mu=w_mu.tolist(), sigma=w_sigma.tolist(), nu=self.model_params.bnn_layer_params[layer_ind].weight.nu
+            )
+
+            name = f"{self._bias_var_name}_{layer_ind}"
+            b_mu = np.mean(trace[name], axis=0)
+            b_sigma = np.std(trace[name], axis=0)
+            self.model_params.bnn_layer_params[layer_ind].bias = StudentTArray(
```
shaharbar1: I know that previously we weren't updating nu. Feels like some historical bug to me. Please consult with @j3rom3c about this. I think nu should also be updated.

pybandits/model.py
```diff
@@ -490,213 +690,311 @@ def check_context_matrix(self, context: np.ndarray):
...
-            # compute mean and std of the coefficients distributions
-            self.alpha.mu = mean(trace["alpha"])
-            self.alpha.sigma = std(trace["alpha"], ddof=1)
-            betas_mu = mean(trace["betas"], axis=1)
-            betas_std = std(trace["betas"], axis=1, ddof=1)
-            self.betas = [
-                StudentT(mu=mu, sigma=sigma, nu=beta.nu) for mu, sigma, beta in zip(betas_mu, betas_std, self.betas)
-            ]
+        for layer_ind in range(len(self.model_params.bnn_layer_params)):
+            name = f"{self._weight_var_name}_{layer_ind}"
```
shaharbar1: This name should be returned from a function, as this convention is used in multiple locations through the code.
That is, you should define _get_weight_name(layer_ind:PositiveInt):...

shaharbar1 (changes requested):

pybandits/model.py
shaharbar1: Furthermore, you should define an init to this model that only takes bnn_layer_params as input, and injects it to bnn_layer_params_init. Otherwise, the usage could be malicious.

pybandits/model.py
ronshiff1: I want to keep this ability to access convergence.  it's going to be useful

pybandits/model.py
ronshiff1: updated docstring

docs/tutorials/cmab.ipynb
ronshiff1: done

pybandits/model.py
ronshiff1: done

pybandits/model.py
ronshiff1: added

pybandits/model.py
ronshiff1: added

pybandits/model.py
ronshiff1: done

pybandits/model.py
ronshiff1: fixed

pybandits/model.py
ronshiff1: done

pybandits/model.py
ronshiff1: done

docs/tutorials/bnn_example.ipynb
ronshiff1: done

pybandits/model.py
ronshiff1: we decided not to touch this in this PR, it might be an improvement in the future

pybandits/model.py
ronshiff1: this is a good question. i don't know. at least in the fit function we don't need it, as we use the logit.
Bernoulli("out", logit_p=logit, observed=bnn_output)

pybandits/model.py
ronshiff1: why do you say it's deterministic?

pybandits/model.py
ronshiff1: done

docs/src/tutorials/bnn_example.ipynb
shaharbar1, line 1: Rename to bnn instead of bnn_example.

docs/src/tutorials/cmab.ipynb
```diff
@@ -186,7 +182,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.5"
+   "version": "3.10.17"
```
shaharbar1: Shouldn't the pre commit hook remove this?

pybandits/model.py
```diff
@@ -490,213 +675,291 @@ def check_context_matrix(self, context: np.ndarray):
...
-        Student's t-distributions of the betas coefficients.
-    update_method : UpdateMethods, defaults to "MCMC"
-        The strategy for computing posterior quantities of the Bayesian models in the update function. Such as Markov
-        chain Monte Carlo ("MCMC") or Variational Inference ("VI"). Check UpdateMethods in pybandits.model for the
-        full list.
-    update_kwargs : Optional[dict], uses default values if not specified
-        Additional arguments to pass to the update method.
-    cost: NonNegativeFloat
-        Cost associated to the Bayesian Logistic Regression model.
+class BayesianLogisticRegressionCC(BayesianLogisticRegression, BayesianNeuralNetworkCC):
```
shaharbar1, line 960: Gr8

tests/test_utils.py
```diff
@@ -24,9 +25,23 @@ def to_temporary_pickle(model: PyBanditsBaseModel):
...
-        return {
-            "alpha": np.random.random(size=(1, self.n_draws)),
-            "betas": np.random.random(size=(self.n_features, self.n_draws)),
-        }
+        sample_dict = {}
+        if self.hidden_dim_list is None:
+            self.hidden_dim_list = []
+        dim_list = [self.n_features] + self.hidden_dim_list + [1]
+        for i in range(len(dim_list) - 1):
+            sample_dict[f"weight_{i}"] = np.random.random(size=(self.n_draws, dim_list[i], dim_list[i + 1]))
```
shaharbar1: _weight_var_name?

docs/src/tutorials/cmab.ipynb
ronshiff1: I will add it to the nbstripout filtering

docs/src/tutorials/bnn_example.ipynb
ronshiff1: done

shaharbar1 (approved):

PR #90, opened 2025-04-23
Fix memory handling in ActionsManager update method

shaharbar1:
* Ensure memory kwargs are correctly stripped of suffix before passing to _update_actions.
* Add test for handling non-matching lengths of actions_memory and rewards_memory in SmabActionsManager.
* Update test_context_memory_matching_dimensions to include other_reward parameter for better test coverage.

### DIFF
--- a/pybandits/actions_manager.py
+++ b/pybandits/actions_manager.py
@@ -236,7 +236,10 @@ class ActionsManager(PyBanditsBaseModel, ABC):
 
                 for action_model in self.actions.values():
                     action_model.reset()
-                self._update_actions(actions_memory, rewards_memory, **memory_kwargs)
+                stripped_memory_kwargs = {
+                    k[: -len(self._memory_parameters_suffix)]: v for k, v in memory_kwargs.items()
+                }
+                self._update_actions(actions_memory, rewards_memory, **stripped_memory_kwargs)
             else:
                 self._update_actions(actions, rewards, **update_kwargs)
         else:
--- a/tests/test_actions_manager.py
+++ b/tests/test_actions_manager.py
@@ -3,10 +3,11 @@ from typing import List, Union
 
 import numpy as np
 import pytest
-from hypothesis import given
+from hypothesis import given, settings
 from hypothesis import strategies as st
 from pytest_mock import MockerFixture
 
+import pybandits
 from pybandits.actions_manager import ActionsManager, CmabActionsManager, SmabActionsManager
 from pybandits.base import ACTION_IDS_PREFIX, ActionId, BinaryReward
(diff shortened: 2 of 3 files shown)


fsimond (approved):

pybandits/actions_manager.py
```diff
@@ -236,7 +236,10 @@ def update(
 
                 for action_model in self.actions.values():
                     action_model.reset()
-                self._update_actions(actions_memory, rewards_memory, **memory_kwargs)
+                stripped_memory_kwargs = {
+                    k[: -len(self._memory_parameters_suffix)]: v for k, v in memory_kwargs.items()
```
adarmiento, line 240: This is a bit dangerous, as you do not have any explicit dependency between the suffix length and the parameter name itself.

If the parameters are few and known in advance, I would evaluate a simple map. If custom logic is required, I would extract this in a method that is easy to monitor and maintain

tests/test_actions_manager.py
```diff
@@ -249,6 +249,36 @@ def test_handles_kwargs_with_no_matching_action_model_attributes(mocker: MockerF
...
+    actions = ["action1"] * data_len
+    rewards = [1] * data_len
+    actions_memory = ["action1"] * memory_len
+    rewards_memory = [other_reward] * memory_len
+    manager.update(actions_memory, rewards_memory)
+    if data_len == memory_len:
+        manager.update(actions, rewards, actions_memory=actions_memory, rewards_memory=rewards_memory)
+    else:
+        with pytest.raises(ValueError):
+            manager.update(actions, rewards, actions_memory=actions_memory, rewards_memory=rewards_memory)
```
adarmiento: This test is generating a side effect, which is not a good practice. 
I would keep 2 different test instead of having a branch condition, in case you are testing happy and failure paths. 
If this is needed, in the with pytest.raises block there should be an assert of some sort to test this specific behavior. 

If you are just injecting setup/cleanup code (I see manager.update() in both if blocks) consider using try/finally instead or a fixture, because right now if the test explode after the      manager.update(actions_memory, rewards_memory) but before the if block, you will have a side effect

pybandits/actions_manager.py
shaharbar1: Hey mate!
I understand the concern but `memory_kwargs` are created based on `kwargs `with a suffix of `self._memory_parameters_suffix `, so I think it's OK here. WDYT?

tests/test_actions_manager.py
shaharbar1: I separated the tests. Thanks for stopping me from being lazy :-)

PR #99, opened 2025-06-30
Fix load_from_old_state when predict_with_proba and predict_actions_randomly exist

ronshiff1:
### Changes: 
*  remove  predict_actions_randomly predict_with_proba and from old state


pybandits/mab.py
shaharbar1, line 1: This change should also be under cmab.

shaharbar1 (changes requested):

shaharbar1 (approved):

PR #101, opened 2025-07-09
Improve model performance

ronshiff1:
### Changes:
 * Improve single prediction speed by ~40%
 * Remove pymc prints

### DIFF
--- a/pybandits/model.py
+++ b/pybandits/model.py
@@ -442,7 +442,6 @@ class BaseBayesianNeuralNetwork(Model, ABC):
     model_params: BnnParams
 
     _logit_var_name: ClassVar[str] = "logit"
-    _prob_var_name: ClassVar[str] = "prob"
     _weight_var_name: ClassVar[str] = "weight"
     _bias_var_name: ClassVar[str] = "bias"
 
@@ -526,6 +525,10 @@ class BaseBayesianNeuralNetwork(Model, ABC):
     @property
     def approx_history(self) -> Optional[np.ndarray]:
         return self._approx_history
+
+    @classmethod
+    def _stable_sigmoid(cls, x):
+        return np.where(x >= 0, 1 / (1 + np.exp(-x)), np.exp(x) / (1 + np.exp(x)))
 
     @classmethod
     def get_layer_params_name(cls, layer_ind: PositiveInt) -> Tuple[str, str]:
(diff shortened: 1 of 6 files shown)


pybandits/model.py
```diff
@@ -53,6 +54,9 @@
 UpdateMethods = Literal["VI", "MCMC"]
 
 
+logging.getLogger("pymc").setLevel(logging.WARNING)
```
shaharbar1: Please have loguru.logger intercept the pymc logging.

pybandits/model.py
```diff
@@ -527,6 +530,10 @@ def arrange_update_kwargs(self):
     def approx_history(self) -> Optional[np.ndarray]:
         return self._approx_history
 
+    @classmethod
+    def stable_sigmoid(cls, x):
```
shaharbar1: make it protected using '_'.

pybandits/model.py
```diff
@@ -653,9 +660,11 @@ def create_model(
             y = np.zeros(len(x), dtype=np.int64)
 
         with PymcModel() as _model:
+            n_samples = len(x)
             # Define data variables using minibatches
             bnn_output = Data("bnn_output", y, mutable=True)
-            if is_predict:
+
+            if is_predict and (n_samples != 1):
```
shaharbar1: Define new variable is_multivariate_predict = is_predict and (n_samples != 1)

tests/test_utils.py
shaharbar1, line 1: Please also add a test for the sample_proba that injects replicated samples of the same context vector and makes sure that the predictions are different.

shaharbar1 (changes requested):

pybandits/model.py
ronshiff1: please explain what do you mean, and what is the purpose

pybandits/model.py
ronshiff1: I'm using this function in:
`def fake_bnn_sample_proba(self, context: np.ndarray, *args, **kwargs) -> List[ProbabilityWeight]:
    n_samples = len(context)
    fake_prediction = FakePrediction(n_samples=n_samples)
    predictions = fake_prediction.sample_prior_predictive()["prior"]
    mock_weighted_sums = predictions[BaseBayesianNeuralNetwork._logit_var_name].values
    mock_probs = BaseBayesianNeuralNetwork._stable_sigmoid(mock_weighted_sums)`
    should I re implement it?

shaharbar1: As long as it's used outside the class only in tests, then yes, make it protected.

pybandits/model.py
shaharbar1: https://github.com/Delgan/loguru/issues/69

shaharbar1 (approved):

PR #111, opened 2025-11-18
Unify quantitative actions interface

shaharbar1:
### Changes:
* Moved `_normalize_field` method to `PyBanditsBaseModel` for default value normalization of optional fields.
* Unified quantitative actions interface to use callables for general form, rather than floats for zooming.
* Improved tests for action selection and quantitative actions, ensuring proper handling of constraints and expected results.


pybandits/utils.py
```diff
@@ -137,3 +146,70 @@ def visualize_via_bokeh(output_path: Optional[str], tabs: List[TabPanel]):
...
+    Parameters
+    ----------
+    quantity_score_func : Callable[[np.ndarray], float]
+        The quantity score function.
+    dimension : PositiveInt
+        The quantity vector dimension.
+    constraint : Optional[List[Callable[[np.ndarray], bool]]]
+        The constraint functions.
+    n_trials : PositiveInt, defaults to 10000
+        The number of otimization trials.
```
AnastasiiaKabeshova: Typo: optimization

pybandits/utils.py
```diff
@@ -137,3 +146,70 @@ def visualize_via_bokeh(output_path: Optional[str], tabs: List[TabPanel]):
...
+    # Only add constraints if they exist
+    if constraints is not None:
+        de_params["constraints"] = constraints
+    result = differential_evolution(**de_params)
+
+    if result.success:
+        return result.x
+    else:
+        logger.warning(f"Optimization failed: {result.message}")
+        return None
```
AnastasiiaKabeshova: When optimization fails, `maximize_by_quantity` returns `None`
Did you check that all callers handle `None` properly?
Whether a more specific exception would be better?

AnastasiiaKabeshova (changes requested):
Great code organization improvement :)

Just one question for myself: Why was Optuna replaced? (Performance, dependency reduction, or other reasons?)

pybandits/base.py
```diff
@@ -45,10 +59,12 @@
 MOProbability = List[Probability]
 MOProbabilityWeight = List[ProbabilityWeight]
 # QuantitativeProbability generalizes probability to include both action quantities and their associated probability
-QuantitativeProbability = Tuple[Tuple[Tuple[Float01, ...], Probability], ...]
-QuantitativeProbabilityWeight = Tuple[Tuple[Tuple[Float01, ...], ProbabilityWeight], ...]
-QuantitativeMOProbability = Tuple[Tuple[Tuple[Float01, ...], List[Probability]], ...]
-QuantitativeMOProbabilityWeight = Tuple[Tuple[Tuple[Float01, ...], List[ProbabilityWeight]], ...]
+QuantitativeProbability = Callable[[np.ndarray], Probability]
```
ronshiff1, line 62: The name of the type is misleading... 
maybe it should be: QuantitativeProbabilityFunction?

pybandits/quantitative_model.py
```diff
@@ -68,6 +78,12 @@ class QuantitativeModel(BaseModelSO, ABC):
     def sample_proba(self, **kwargs) -> List[QuantitativeProbability]:
```
ronshiff1, line 79: Please explain why list. For Multi objective?

ronshiff1: this is for a list of samples?

ronshiff1: If I understand correctly - it's for different segments?
I think that it is very specific to the zooming algorithm, and it should be function which receives an input, and outputs one function, and not a list
the zooming model should create this function and handle the segmentation internally (I also commented on the first sample_proba implementation)

pybandits/quantitative_model.py
```diff
@@ -503,17 +519,36 @@ def _generate_initial_segments(cls, dimension: PositiveInt) -> List[Tuple[Tuple[
 
     def sample_proba(self, **kwargs) -> List[QuantitativeProbability]:
```
ronshiff1, line 523: as I wrote in my last comment about the abstract sample_proba - I think that the structure should be as follows (and for BNN it will be similar)
pseudo code:
def sample_proba(self, **kwargs) -> QuantitativeProbability:
list_sampled_values = sample_values_for_each_segment()
quantitaive_action_to_reward_func = create_reward_func_from_sampled_values(list_sampled_values)

pybandits/quantitative_model.py
shaharbar1: The list is not because of zooming. It returns a list whose length is as the number of samples.

shaharbar1: I fixed the docstring to make it explicit.

pybandits/base.py
shaharbar1: @ronshiff1 Maybe ContinuousProbability? I just didn't want it to be too long.

pybandits/utils.py
shaharbar1: Thx!

pybandits/utils.py
shaharbar1: Good point, I added an appropriate exception handling.

pybandits/quantitative_model.py
```diff
@@ -926,6 +998,49 @@ def _init_base_model(self):
...
+                sample_idx: NonNegativeInt, output_index: NonNegativeInt
+            ) -> Union[QuantitativeProbability, QuantitativeWeight]:
+                def output_function(quantity: np.ndarray) -> Union[Probability, float]:  # Probability or weight
+                    """
+                    Evaluate output at the given quantity.
+                    """
+                    for segment in segment_probabilities.keys():
+                        if quantity in segment:
+                            segment_probas_for_segment = segment_probabilities[segment]
+                            return segment_probas_for_segment[sample_idx][output_index]  # Probability or weight
```
ronshiff1, line 1044: Why not directly return segment_probas_for_segment[sample_idx] which is tuple?
(so def create_probability_or_weight_function(
                sample_idx: NonNegativeInt
            ) -> **Tuple**[QuantitativeProbability, QuantitativeWeight]:

shaharbar1: Then in the cmab class you would need to separate them to list of probabilities and list of weights.
So I preferred to do it in advance.

pybandits/base.py
```diff
@@ -190,6 +206,28 @@ def _get_field_type(cls, key: str) -> Any:
             annotation = get_args(annotation)
         return annotation
 
+    @classmethod
+    def _normalize_field(cls, v: Any, field_name: str) -> Any:
```
ronshiff1, line 210: why do we need this?

pybandits/quantitative_model.py
ronshiff1, line 616: Will this work when quantities is float?
[segment for value in quantities....

ronshiff1, line 897: this:
quantities: Optional[List[Union[float, List[float], None]]],
        rewards: Union[List[BinaryReward], List[List[BinaryReward]]],
doesn't conform with the ZoomingModel _quantitative_update:
  def _quantitative_update(self, quantities: List[Union[float, np.ndarray]], rewards: List[BinaryReward], **kwargs):

ronshiff1 (approved):

pybandits/base.py
shaharbar1: I use it in validators of "pre" behavior to get the default value of an ungiven attribute

pybandits/quantitative_model.py
shaharbar1: Wow! Nice catch. Fixed to: Optional[List[Union[float, List[float], None]]]

pybandits/mab.py
ronshiff1, line 315: say, shouldnt it be 
-> Union[ActionId, QuantitativeActionId]?

docs/src/tutorials/cmab_zooming.ipynb
ronshiff1, line 236: isn't current_context[0] a bug?
it uses the first context for all samples

PR #115, opened 2026-01-06
BNN with Normal Priors

AnastasiiaKabeshova:
### Changes:
* Added Normal distribution support as a Student-t distribution with nu parameter that approaches to infinity
* Update tutorial notebook for BNN enhancements
* Tests added or updated

### DIFF
--- a/pybandits/model.py
+++ b/pybandits/model.py
@@ -31,9 +31,10 @@ from numpy import sqrt
 from numpy.typing import ArrayLike
 from pymc import Bernoulli, Data, Deterministic, Minibatch, fit, math, sample
 from pymc import Model as PymcModel
+from pymc import Normal as PymcNormal
 from pymc import StudentT as PymcStudentT
 from scipy.special import erf
-from scipy.stats import t
+from scipy.stats import norm, t
 from typing_extensions import Self
 
 from pybandits.base import BinaryReward, MOProbability, Probability, ProbabilityWeight, PyBanditsBaseModel
@@ -278,51 +279,122 @@ class BetaMOCC(BaseBetaMO, ModelCC):
     """
 
 
-class StudentTArray(PyBanditsBaseModel):
+class BaseLocationScaleArray(PyBanditsBaseModel, ABC):
     """
-    A class representing an array of Student's t-distributions with parameters `mu`, `sigma`, and `nu`.
-    A specific element (e.g, a single parameter of a layer) distribution is defined by the the corresponding elements in the lists.
-    The mean values are represented by `mu`, the scale (standard deviation) values by `sigma`, and the degrees of freedom by `nu`.
-
+    Abstract base class for location-scale distribution arrays used in Bayesian Neural Networks.
 
     Parameters
     ----------
     mu : Union[List[float], List[List[float]]]
-        The mean values of the Student's t-distributions. Can be a 1D (for the layer bias term) or 2D list (for the layer weight term).
+        The mean values of the distributions. Can be a 1D (for the layer bias term) or 2D list (for the layer weight term).
     sigma : Union[List[NonNegativeFloat], List[List[NonNegativeFloat]]]
-        The scale (standard deviation) values of the Student's t-distributions. Must be non-negative.
-        Can be a 1D or 2D list.
-    nu : Union[List[PositiveFloat], List[List[PositiveFloat]]]
-        The degrees of freedom of the Student's t-distributions. Must be positive.
+        The scale (standard deviation) values of the distributions. Must be non-negative.
         Can be a 1D or 2D list.
     """
 
     mu: Union[List[float], List[List[float]]]
     sigma: Union[List[NonNegativeFloat], List[List[NonNegativeFloat]]]
-    nu: Union[List[PositiveFloat], List[List[PositiveFloat]]]
 
     _mu_array: np.ndarray = PrivateAttr()
     _sigma_array: np.ndarray = PrivateAttr()
-    _nu_array: np.ndarray = PrivateAttr()
     _params: Dict[str, np.ndarray] = PrivateAttr()
+    _pymc_class: ClassVar[type]
 
-    def __eq__(self, other: Any) -> bool:
-        if not isinstance(other, StudentTArray):
-            return False
-        return (
-            np.all(self._mu_array == other._mu_array)
-            and np.all(self._sigma_array == other._sigma_array)
-            and np.all(self._nu_array == other._nu_array)
-        )
+    def to_pymc_distribution(
+        self, name: str, shape: Tuple[PositiveInt, ...], initval: Literal["prior"] = "prior"
+    ) -> Union[PymcStudentT, PymcNormal]:
+        """
+        Create a PyMC distribution from this prior distribution array.
+
+        Parameters
+        ----------
+        name : str
+            Name for the PyMC random variable.
+        shape : Tuple[PositiveInt, ...]
+            Shape of the distribution array.
+        initval : Literal["prior"], optional
+            Initialization value strategy, by default "prior".
+
+        Returns
+        -------
+        Union[PymcStudentT, PymcNormal]
+            A PyMC distribution instance.
+        """
+        if self._pymc_class is None:
+            raise NotImplementedError(
+                f"{self.__class__.__name__} must define _pymc_class ClassVar to specify the PyMC distribution class."
+            )
+        return self._pymc_class(name=name, shape=shape, **self.params, initval=initval)
+
+    def with_dist_parameters(self, **kwargs) -> "BaseLocationScaleArray":
+        """
+        Create a new instance with updated distribution parameters.
+
+        Parameters
+        ----------
+        **kwargs
+            Parameters to update (e.g., `mu`, `sigma`, `nu` for StudentTArray).
+            If empty, returns self unchanged.
+
+        Returns
+        -------
+        BaseLocationScaleArray
+            A new instance with the updated parameters.
+        """
+        if not kwargs:
+            return self
+
+        # Convert to dict, update with new parameters, and validate to create new instance
+        updated_dict = self._apply_version_adjusted_method("model_dump", "dict")
+        updated_dict.update(kwargs)
+        if pydantic_version == PYDANTIC_VERSION_1:
+            return self.__class__(**updated_dict)
+        elif pydantic_version == PYDANTIC_VERSION_2:
+            return self.__class__.model_validate(updated_dict)
+        else:
+            raise ValueError(f"Unsupported pydantic version: {pydantic_version}")
+
+    @abstractmethod
+    def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
+        """
+        Sample random variates from this distribution.
+
+        Parameters
+        ----------
+        size : Tuple[int, ...]
+            Shape of the output array.
+
+        Returns
+        -------
+        np.ndarray
+            Array of sampled values.
+        """
+        pass
 
     @staticmethod
-    def maybe_convert_list_to_array(input_list: Union[List[float], List[List[float]]]) -> bool:
+    def maybe_convert_list_to_array(input_list: Union[List[float], List[List[float]]]) -> np.ndarray:
+        """
+        Convert a list or list of lists to a numpy array.
+
+        Parameters
+        ----------
+        input_list : Union[List[float], List[List[float]]]
+            Input list to convert.
+
+        Returns
+        -------
+        np.ndarray
+            Converted numpy array.
+
+        Raises
+        ------
+        ValueError
+            If the input list is not a valid 1D or 2D list.
+        """
         if len(input_list) == 0:
             is_valid_input = False
-
         elif not isinstance(input_list[0], list):
             is_valid_input = True
-
         else:
             first_length = len(input_list[0])
             is_valid_input = all(
@@ -337,27 +409,172 @@ class StudentTArray(PyBanditsBaseModel):
     @model_validator(mode="before")
     @classmethod
     def validate_input_shapes(cls, values):
-        mu_input = values.get("mu")
-        sigma_input = values.get("sigma")
-        nu_input = values.get("nu")
+        """
+        Validate that all array-like parameters have the same shape.
 
-        mu_arr = cls.maybe_convert_list_to_array(mu_input)
-        sigma_arr = cls.maybe_convert_list_to_array(sigma_input)
-        nu_arr = cls.maybe_convert_list_to_array(nu_input)
+        Parameters
+        ----------
+        values : dict or BaseLocationScaleArray instance
+            Dictionary of field values or an already-instantiated object.
 
-        if (mu_arr.shape != sigma_arr.shape) or (mu_arr.shape != nu_arr.shape):
-            raise ValueError(
-                f"mu, sigma, and nu must have the same shape, "
-                f"but are {mu_arr.shape}, {sigma_arr.shape}, and {nu_arr.shape}, respectively."
-            )
+        Returns
+        -------
+        dict or BaseLocationScaleArray instance
+            Validated values dictionary or the object itself if already instantiated.
 
-        if any(dim_len == 0 for dim_len in mu_arr.shape):
-            raise ValueError("mu, sigma, and nu must have at least one element in every dimension.")
+        Raises
+        ------
+        ValueError
+            If array-like parameters have different shapes or empty dimensions.
+        """
+        # If values is already an instance of this class or a subclass, return it as-is
+        if isinstance(values, cls):
+            return values
 
-        for key, value in zip(["mu", "sigma", "nu"], [mu_input, sigma_input, nu_input]):
-            if isinstance(value, np.ndarray):
-                values[key] = value.tolist()
+        # If values is not a dict, it might be another type - return as-is for Pydantic to handle
+        if not isinstance(values, dict):
+            return values
+
+        # Find all array-like values (lists or numpy arrays) that need shape validation
+        array_like_keys = []
+        array_like_values = []
+        arrays = []
+
+        for key, value in values.items():
+            # Skip None values and non-array-like types
+            if value is None:
+                continue
+            if isinstance(value, (list, np.ndarray)):
+                array_like_keys.append(key)
+                array_like_values.append(value)
+                # Convert to array for shape comparison
+                if isinstance(value, list):
+                    arr = cls.maybe_convert_list_to_array(value)
+                else:
+                    arr = value
+                arrays.append(arr)
+
+        # If we have array-like values, validate they all have the same shape
+        if arrays:
+            reference_shape = arrays[0].shape
+
+            # Check all arrays have the same shape
+            for key, arr in zip(array_like_keys, arrays):
+                if arr.shape != reference_shape:
+                    raise ValueError(
+                        f"All array-like parameters must have the same shape, but {key} has shape {arr.shape} "
+                        f"while the reference shape is {reference_shape}."
+                    )
+
+            # Check no empty dimensions
+            if any(dim_len == 0 for dim_len in reference_shape):
+                param_names = ", ".join(array_like_keys)
+                raise ValueError(
+                    f"All array-like parameters ({param_names}) must have at least one element in every dimension."
+                )
+
+            # Convert numpy arrays back to lists for Pydantic
+            for key, value in zip(array_like_keys, array_like_values):
+                if isinstance(value, np.ndarray):
+                    values[key] = value.tolist()
+
         return values
+
+    def model_post_init(self, __context: Any) -> None:
+        """
+        Initialize private numpy array attributes by converting lists to arrays once at initialization.
+
+        Parameters
+        ----------
+        __context : Any
+            Pydantic context (unused).
+        """
+        self._mu_array = np.array(self.mu)
+        self._sigma_array = np.array(self.sigma)
+        self._params = dict(mu=self._mu_array, sigma=self._sigma_array)
+
+    @property
+    def shape(self) -> Tuple[PositiveInt, ...]:
+        """
+        Get the shape of the mu array.
+
+        Returns
+        -------
+        Tuple[PositiveInt, ...]
+            The shape of the mu array.
+        """
+        return self._mu_array.shape
+
+    @property
+    def params(self) -> Dict[str, np.ndarray]:
+        """
+        Get the parameters as a dictionary of numpy arrays.
+
+        Returns
+        -------
+        Dict[str, np.ndarray]
+            Dictionary containing 'mu' and 'sigma' as numpy arrays.
+        """
+        return self._params
+
+    def __eq__(self, other: Any) -> bool:
+        """
+        Check equality with another distribution array.
+
+        Parameters
+        ----------
+        other : Any
+            Other object to compare with.
+
+        Returns
+        -------
+        bool
+            True if distributions are equal, False otherwise.
+        """
+        if not isinstance(other, BaseLocationScaleArray):
+            return False
+        return (
+            np.all(self._mu_array == other._mu_array)
+            and np.all(self._sigma_array == other._sigma_array)
+            and type(self) is type(other)
+        )
+
+
+class StudentTArray(BaseLocationScaleArray):
+    """
+    A class representing an array of Student's t-distributions with parameters `mu`, `sigma`, and `nu`.
+    A specific element (e.g, a single parameter of a layer) distribution is defined by the the corresponding elements in the lists.
+    The mean values are represented by `mu`, the scale (standard deviation) values by `sigma`, and the degrees of freedom by `nu`.
+
+    Parameters
+    ----------
+    mu : Union[List[float], List[List[float]]]
+        The mean values of the Student's t-distributions. Can be a 1D (for the layer bias term) or 2D list (for the layer weight term).
+    sigma : Union[List[NonNegativeFloat], List[List[NonNegativeFloat]]]
+        The scale (standard deviation) values of the Student's t-distributions. Must be non-negative.
+        Can be a 1D or 2D list.
+    nu : Union[List[PositiveFloat], List[List[PositiveFloat]]]
+        The degrees of freedom of the Student's t-distributions. Must be positive.
+        Can be a 1D or 2D list.
+    """
+
+    nu: Union[List[PositiveFloat], List[List[PositiveFloat]]]
+
+    _nu_array: np.ndarray = PrivateAttr()
+    _pymc_class: ClassVar[type] = PymcStudentT
+
+    @model_validator(mode="before")
+    @classmethod
+    def validate_input_shapes(cls, values):
+        # The parent class method is now generic and handles all array-like parameters
+        # including mu, sigma, and nu, so we can just call it directly
+        return super().validate_input_shapes(values)
+
+    def __eq__(self, other: Any) -> bool:
+        """Check equality including nu parameter."""
+        if not isinstance(other, StudentTArray):
+            return False
+        return super().__eq__(other) and np.all(self._nu_array == other._nu_array)
 
     @classmethod
     def cold_start(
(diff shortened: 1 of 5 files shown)


pybandits/model.py
```diff
@@ -80,7 +85,13 @@ def _numpy_gelu(x: np.ndarray) -> np.ndarray:
...
 def _stable_sigmoid(x):
     """Stable sigmoid activation function for NumPy."""
-    return np.where(x >= 0, 1 / (1 + np.exp(-x)), np.exp(x) / (1 + np.exp(x)))
+    x_clipped = np.clip(x, -500, 500)
+    return np.where(x_clipped >= 0, 1 / (1 + np.exp(-x_clipped)), np.exp(x_clipped) / (1 + np.exp(x_clipped)))
+
+
+# Threshold for degrees of freedom above which Student-t is treated as Normal
+# Student-t with nu → ∞ converges to Normal distribution
+_NORMAL_NU_THRESHOLD: float = 1e6
```
shaharbar1: Can't we specifically use float("inf") for Normal distribution?

pybandits/model.py
```diff
@@ -80,7 +85,13 @@ def _numpy_gelu(x: np.ndarray) -> np.ndarray:
 
 def _stable_sigmoid(x):
     """Stable sigmoid activation function for NumPy."""
-    return np.where(x >= 0, 1 / (1 + np.exp(-x)), np.exp(x) / (1 + np.exp(x)))
+    x_clipped = np.clip(x, -500, 500)
```
shaharbar1: Why do we need this change?
Also please rename to _numpy_sigmoid.

pybandits/model.py
```diff
@@ -841,26 +944,49 @@ def sample_proba(self, context: np.ndarray) -> List[ProbabilityWeight]:
...
-            w = t.rvs(
-                w_params["nu"],
-                loc=w_params["mu"],
-                scale=w_params["sigma"],
-                size=(n_samples, len(w_params["nu"]), len(w_params["nu"][0])),
-            )
-            b = t.rvs(
-                b_params["nu"], loc=b_params["mu"], scale=b_params["sigma"], size=(n_samples, len(b_params["nu"]))
-            )
+            if w_is_normal:
```
shaharbar1: Isn't it possible that some weights are normal and some are standard student-t?

shaharbar1 (changes requested):

pybandits/model.py
AnastasiiaKabeshova: Because of this warring message : 
> RuntimeWarning: overflow encountered in exp

ronshiff1: when running what?

pybandits/model.py
AnastasiiaKabeshova: yes we can, will do

pybandits/model.py
```diff
@@ -788,12 +864,39 @@ def create_update_model(
                 output_dim = w_shape[1]
 
                 # For training, use shared weights and biases
-                w = PymcStudentT(
-                    name=weight_layer_params_name, shape=w_shape, **layer_params.weight.params, initval="prior"
-                )
-                b = PymcStudentT(
-                    name=bias_layer_params_name, shape=b_shape, **layer_params.bias.params, initval="prior"
-                )
+                # Use Normal if nu is effectively infinite (>= threshold), otherwise use Student-t
```
ronshiff1: - why do we touch the update process? it shouldn't be affected
- If you do want to touch it - I would create a prior property, and handle all the cases there - and not inside this function. the code will become unreadable

pybandits/model.py
ronshiff1: same comment as in the update - I suggest to remove the "if"'s from this function and create a property

ronshiff1 (changes requested):

pybandits/model.py
AnastasiiaKabeshova: renaming is done in the PR of Shahar

pybandits/model.py
AnastasiiaKabeshova: outdated

AnastasiiaKabeshova: in the new version yes

pybandits/model.py
AnastasiiaKabeshova: updated

pybandits/model.py
```diff
@@ -278,51 +279,90 @@ class BetaMOCC(BaseBetaMO, ModelCC):
     """
 
 
-class StudentTArray(PyBanditsBaseModel):
+class PriorDistributionArray(PyBanditsBaseModel, ABC):
```
shaharbar1: I would name it BaseLocationScaleDistributionArray.

pybandits/model.py
```diff
@@ -404,17 +584,159 @@ def shape(self) -> Tuple[PositiveInt, ...]:
...
 
         Returns
         -------
-        Dict[str, np.ndarray]
-            Dictionary containing 'mu', 'sigma', and 'nu' as numpy arrays.
+        PymcStudentT
+            A PyMC StudentT distribution instance.
         """
-        return self._params
+        return PymcStudentT(name=name, shape=shape, **self.params, initval=initval)
```
shaharbar1: you can define a ClassVar in the father class:
_pymc_class : ClassVar
Taking the values PymcStudentT and PymcNormal and use to make to_pymc_distribution non abstract in the father class.

pybandits/model.py
```diff
@@ -404,17 +584,159 @@ def shape(self) -> Tuple[PositiveInt, ...]:
...
-        Get the parameters as a dictionary of numpy arrays.
+        Create a PyMC StudentT distribution from this prior distribution array.
+
+        Parameters
+        ----------
+        name : str
+            Name for the PyMC random variable.
+        shape : Tuple[PositiveInt, ...]
+            Shape of the distribution array.
+        initval : str, optional
```
shaharbar1: This should be Literal, no string

pybandits/model.py
```diff
@@ -337,28 +377,169 @@ def maybe_convert_list_to_array(input_list: Union[List[float], List[List[float]]
...
+        dict or PriorDistributionArray instance
+            Validated values dictionary or the object itself if already instantiated.
+
+        Raises
+        ------
+        ValueError
+            If mu and sigma have different shapes or empty dimensions.
+        """
+        # If values is already an instance of this class or a subclass, return it as-is
+        if isinstance(values, cls):
```
shaharbar1, line 431: Why is this addition needed?

pybandits/model.py
```diff
@@ -423,14 +745,16 @@ class BnnLayerParams(PyBanditsBaseModel):
...
+        The weight parameter of the BNN layer, represented as a PriorDistributionArray
+        (either StudentTArray or NormalArray).
+    bias : PriorDistribution
+        The bias parameter of the BNN layer, represented as a PriorDistributionArray
+        (either StudentTArray or NormalArray).
     """
 
-    weight: StudentTArray
-    bias: StudentTArray
+    weight: PriorDistribution
```
shaharbar1: Why not use the base class here instead of the type hint?

pybandits/model.py
```diff
@@ -937,12 +1272,20 @@ def _update(self, context: np.ndarray, rewards: List[BinaryReward]):
...
-                    )
-                    layer_params.bias = StudentTArray(
-                        mu=b_mu, sigma=b_sigma, nu=self.model_params.bnn_layer_params[layer_ind].bias.nu
-                    )
+
+                    # Preserve the distribution type (StudentTArray or NormalArray)
+                    original_weight = self.model_params.bnn_layer_params[layer_ind].weight
+                    original_bias = self.model_params.bnn_layer_params[layer_ind].bias
+
+                    if isinstance(original_weight, StudentTArray):
```
shaharbar1: Instead of this, implement a "with_dist_parameters(**kwargs)" in the father class, that takes new parameters and returns new instance. It is easy with pydantic, you can find other examples in the code of pybandits, look for "with_"

shaharbar1: Anyhow, this way you need to pass nu, as it always remains the same

pybandits/model.py
```diff
@@ -996,6 +1351,9 @@ def cold_start(
             Additional keyword arguments for the update method. Default is None.
         dist_params_init : Optional[Dict[str, float]], optional
```
shaharbar1, line 1361: Split it into two parameters, one for the type of distribution, and one for the parameters like before. Make the type parameter Literal["normal", "studentt"].
Add a dict class attribute:
_distribution_mapping = {"normal": NormalArray, "studentt":StuddentTArray} and use it to transform to the correct class instead of the if structure.

pybandits/model.py
```diff
@@ -1012,8 +1370,25 @@ def cold_start(
...
 
+        # Copy dist_params_init to avoid mutating the original
+        dist_params_init = dist_params_init.copy()
+
+        dist_type = dist_params_init.pop("dist_type", "studentt")
+
+        if dist_type == "normal":
+            dist_class = NormalArray
+            # Remove 'nu' from dist_params_init if present (Normal doesn't use it)
+            dist_params_init = {k: v for k, v in dist_params_init.items() if k != "nu"}
```
shaharbar1: Not needed, if someone inputs nu for normal distribution, we want pydantic to fail it during initaliztion of the Array class

shaharbar1 (changes requested):

pybandits/model.py
AnastasiiaKabeshova: I find it too Long (harder to read/write)
is LocationScaleArray will convene you?

pybandits/model.py
AnastasiiaKabeshova: The check handles cases where Pydantic receives already-instantiated objects instead of dictionaries.

For example, in create_model_params(), BnnLayerParams is created with already-instantiated objects:
`w_param = dist_class.cold_start(shape=(input_dim, output_dim), **dist_params_init)  # Returns StudentTArray or NormalArrayb_param = dist_class.cold_start(shape=output_dim, **dist_params_init)  # Returns StudentTArray or NormalArraylayer_params_init.append(BnnLayerParams(weight=w_param, bias=b_param))  # Passes objects, not dicts!`
When Pydantic processes BnnLayerParams(weight=w_param, bias=b_param), it calls validate_input_shapes() for the weight and bias fields. Without the check, the validator would try:
mu_input = values.get("mu")  # ❌ Fails! 'StudentTArray' object has no attribute 'get'
This fails because .get() is a dictionary method, not an object method.

pybandits/model.py
AnastasiiaKabeshova: please check the new implementation

pybandits/model.py
shaharbar1: BaseLocationScaleArray would be best. I prefer adding "Base" to indicate abstract class.

tests/test_model.py
```diff
@@ -554,6 +706,63 @@ def update(context: np.ndarray, rewards: list):
         bnn.update(context=context, rewards=rewards[1:])
 
 
+def test_bnn_create_update_model_with_normal_priors():
```
shaharbar1: I like this new pytest!
But please make it available also for studentT :-)

tests/test_model.py
```diff
@@ -530,16 +665,33 @@ def update(context: np.ndarray, rewards: list):
                 assert np.all(np.array(layer_w[param]) == init_params[param])
                 assert np.all(np.array(layer_b[param]) == init_params[param])
 
+            # Verify distribution type
+            if dist_type == "normal":
+                assert isinstance(bnn.model_params.bnn_layer_params[layer_ind].weight, NormalArray)
```
shaharbar1: Change this code snippet:

`expected_array = NormalArray if dist_type == "normal" else StudentTArray
assert isinstance(bnn.model_params.bnn_layer_params[layer_ind].weight, expected_array)
assert isinstance(bnn.model_params.bnn_layer_params[layer_ind].bias, expected_array)
`

tests/test_model.py
```diff
@@ -530,16 +665,33 @@ def update(context: np.ndarray, rewards: list):
...
+        # mu and sigma are updated:
         for layer_ind in range(len(dim_list)):
             layer_w = bnn.model_params.bnn_layer_params[layer_ind].weight.params
             layer_b = bnn.model_params.bnn_layer_params[layer_ind].bias.params
             for param in ["mu", "sigma"]:
                 assert np.all(np.array(layer_w[param]) != init_params[param])
                 assert np.all(np.array(layer_b[param]) != init_params[param])
 
+        # Verify distribution type is preserved after update
+        for layer_ind in range(len(dim_list)):
```
shaharbar1: Change also this

tests/test_model.py
```diff
@@ -402,16 +512,33 @@ def update(context: np.ndarray, rewards: list):
...
+            if dist_type == "normal":
+                assert isinstance(bnn.model_params.bnn_layer_params[layer_ind].weight, NormalArray)
```
shaharbar1: Change also this like in my comment on the update pytest

tests/test_model.py
```diff
@@ -402,16 +512,33 @@ def update(context: np.ndarray, rewards: list):
                 assert np.all(np.array(layer_w[param]) == init_params[param])
                 assert np.all(np.array(layer_b[param]) == init_params[param])
 
+            # Verify distribution type
+            if dist_type == "normal":
+                assert isinstance(bnn.model_params.bnn_layer_params[layer_ind].weight, NormalArray)
```
shaharbar1: Change also this like in my comment on the update pytest

shaharbar1 (changes requested):

pybandits/model.py
```diff
@@ -278,51 +279,129 @@ class BetaMOCC(BaseBetaMO, ModelCC):
...
 
     mu: Union[List[float], List[List[float]]]
     sigma: Union[List[NonNegativeFloat], List[List[NonNegativeFloat]]]
-    nu: Union[List[PositiveFloat], List[List[PositiveFloat]]]
 
     _mu_array: np.ndarray = PrivateAttr()
     _sigma_array: np.ndarray = PrivateAttr()
-    _nu_array: np.ndarray = PrivateAttr()
     _params: Dict[str, np.ndarray] = PrivateAttr()
+    _pymc_class: ClassVar[type] = None  # Must be set by subclasses
```
shaharbar1: Just
_pymc_class: ClassVar

No need to make it None

pybandits/model.py
```diff
@@ -278,51 +279,129 @@ class BetaMOCC(BaseBetaMO, ModelCC):
...
+        """
+        if self._pymc_class is None:
+            raise NotImplementedError(
+                f"{self.__class__.__name__} must define _pymc_class ClassVar to specify the PyMC distribution class."
+            )
+        return self._pymc_class(name=name, shape=shape, **self.params, initval=initval)
+
+    def with_dist_parameters(
+        self,
+        mu: Optional[Union[List[float], List[List[float]]]] = None,
```
shaharbar1: I think mu and sigma are redundant here. WDYT?
We can use kwargs directly and skip the redundant if statements.

pybandits/model.py
```diff
@@ -337,28 +416,170 @@ def maybe_convert_list_to_array(input_list: Union[List[float], List[List[float]]
...
+        """
+        # If values is already an instance of this class or a subclass, return it as-is
+        if isinstance(values, cls):
+            return values
+
+        # If values is not a dict, it might be another type - return as-is for Pydantic to handle
+        if not isinstance(values, dict):
+            return values
+
         mu_input = values.get("mu")
```
shaharbar1: Make this function generic. That is, instead of relying on "mu", "sigma" and adding "nu" later, work directly according to dict keys. Compare their shapes, maybe_convert_list_to_array, etc.

pybandits/model.py
```diff
@@ -387,10 +608,9 @@ def model_post_init(self, __context: Any) -> None:
         __context : Any
             Pydantic context (unused).
         """
-        self._mu_array = np.array(self.mu)
-        self._sigma_array = np.array(self.sigma)
+        super().model_post_init(__context)
         self._nu_array = np.array(self.nu)
-        self._params = dict(mu=self._mu_array, sigma=self._sigma_array, nu=self._nu_array)
+        self._params["nu"] = self._nu_array
```
shaharbar1, line 609: Gr8!

pybandits/model.py
```diff
@@ -423,14 +738,48 @@ class BnnLayerParams(PyBanditsBaseModel):
...
+        This is needed because BaseLocationScaleArray is abstract and cannot be instantiated directly.
+        """
+        if isinstance(values, dict):
+            # Handle weight
+            if "weight" in values:
+                weight_val = values["weight"]
+                # Convert dict to appropriate concrete class
+                if isinstance(weight_val, dict):
+                    # Determine class based on presence of 'nu' key
+                    if "nu" in weight_val:
```
shaharbar1: I don't think that this validation is needed. Pydantic should know which class to use.

shaharbar1: OK, my bad, it won't, for it to be able to do so, you need
weight: Union[NormalArray, StudentTArray]
bias: Union[StudentTArray, NormalArray]

I made you change it in my first PR comments, sorry, my bad

pybandits/cmab.py
```diff
@@ -253,9 +253,43 @@ def update_old_state(
                 weight_init = StudentTArray.cold_start(shape=(len(mu_list), 1))
                 layer_params_init = BnnLayerParams(weight=weight_init, bias=bias_init)
 
+                # model_dump() should include all fields by default, including "nu" for StudentTArray
                 model_params = BnnParams(
                     bnn_layer_params=[layer_params], bnn_layer_params_init=[layer_params_init]
                 )._apply_version_adjusted_method("model_dump", "dict")
+
+                if "bnn_layer_params" in model_params and len(model_params["bnn_layer_params"]) > 0:
```
shaharbar1: Do the fix I suggested to revert under the BNN class and you won't need this fix under update_old_state

pybandits/model.py
```diff
@@ -653,7 +1020,11 @@ def get_layer_params_name(cls, layer_ind: PositiveInt) -> Tuple[str, str]:
 
     @classmethod
     def create_model_params(
-        cls, n_features: PositiveInt, hidden_dim_list: List[PositiveInt], **dist_params_init
+        cls,
+        n_features: PositiveInt,
+        hidden_dim_list: List[PositiveInt],
+        dist_class: type = StudentTArray,
```
shaharbar1: make it type[BaseLocationScaleArray]

pybandits/model.py
```diff
@@ -502,6 +868,7 @@ class BaseBayesianNeuralNetwork(Model, ABC):
     _weight_var_name: ClassVar[str] = "weight"
     _bias_var_name: ClassVar[str] = "bias"
     _vi_update_params: ClassVar[list] = ["optimizer_type", "optimizer_kwargs", "batch_size"]
+    _distribution_mapping: ClassVar[Dict[str, type]] = {"normal": NormalArray, "studentt": StudentTArray}
```
shaharbar1, line 833: Gr8, thanks!

pybandits/model.py
```diff
@@ -841,26 +1213,17 @@ def sample_proba(self, context: np.ndarray) -> List[ProbabilityWeight]:
...
-                loc=w_params["mu"],
-                scale=w_params["sigma"],
-                size=(n_samples, len(w_params["nu"]), len(w_params["nu"][0])),
-            )
-            b = t.rvs(
-                b_params["nu"], loc=b_params["mu"], scale=b_params["sigma"], size=(n_samples, len(b_params["nu"]))
-            )
+            # Sample weights and biases using the distribution's sample_rvs method
+            w = layer_params.weight.sample_rvs(size=(n_samples, input_dim, output_dim))
+            b = layer_params.bias.sample_rvs(size=(n_samples, output_dim))
```
shaharbar1, line 1188: Very nice!

pybandits/model.py
```diff
@@ -887,6 +1250,52 @@ def sample_proba(self, context: np.ndarray) -> List[ProbabilityWeight]:
...
+
+        Returns
+        -------
+        BnnLayerParams
+            New layer parameters with updated values.
+        """
+        # Create new distribution objects with updated parameters, preserving the type
+        # For StudentTArray, preserve nu; for NormalArray, don't include nu
+        if isinstance(original_weight, StudentTArray):
+            updated_weight = StudentTArray(mu=w_mu.tolist(), sigma=w_sigma.tolist(), nu=original_weight.nu)
```
shaharbar1: Here you can use
updated_weight = original_weight.with_dist_parameters(...)

pybandits/model.py
```diff
@@ -937,35 +1345,34 @@ def _update(self, context: np.ndarray, rewards: List[BinaryReward]):
...
-                    layer_params.bias = StudentTArray(
-                        mu=b_mu, sigma=b_sigma, nu=self.model_params.bnn_layer_params[layer_ind].bias.nu
+
+                    updated_layer_params = self._create_updated_layer_params(
+                        w_mu, w_sigma, b_mu, b_sigma, layer_params.weight, layer_params.bias
                     )
-                    self.model_params.bnn_layer_params[layer_ind] = layer_params
+                    updated_layer_params_list.append(updated_layer_params)
+
+                self.model_params.bnn_layer_params = updated_layer_params_list
```
shaharbar1: Take this line out of the if and else, it is joint for VI and MCMC

pybandits/model.py
```diff
@@ -994,8 +1402,12 @@ def cold_start(
             Method to update the network, either "MCMC" or "VI". Default is "MCMC".
         update_kwargs : Optional[dict], optional
             Additional keyword arguments for the update method. Default is None.
-        dist_params_init : Optional[Dict[str, float]], optional
+        dist_type : Literal["normal", "studentt"], optional
```
shaharbar1: It is not optional, it just has a default value

pybandits/model.py
```diff
@@ -994,8 +1402,12 @@ def cold_start(
...
+            Type of distribution to use for priors. Default is "studentt".
+        dist_params : Optional[Dict[str, float]], optional
```
shaharbar1: Keep it dist_params_init. Less work for MLEs...

pybandits/model.py
```diff
@@ -1009,11 +1421,23 @@ def cold_start(
             An instance of the Bayesian Neural Network initialized with the specified parameters.
         """
 
-        if dist_params_init is None:
-            dist_params_init = {}
+        if dist_params is None:
+            dist_params = {}
+
+        # Copy dist_params to avoid mutating the original
+        dist_params = dist_params.copy()
```
shaharbar1: Why do we care about mutation?

pybandits/model.py
```diff
@@ -1009,11 +1421,23 @@ def cold_start(
...
+
+        # Get distribution class from mapping
+        if dist_type not in cls._distribution_mapping:
+            raise ValueError(
+                f"Invalid dist_type: {dist_type}. Must be one of {list(cls._distribution_mapping.keys())}. "
+                f"Example: dist_type='normal', dist_params={{'mu': 0, 'sigma': 1}}"
+            )
+
+        dist_class = cls._distribution_mapping[dist_type]
```
shaharbar1, line 1387: Gr8!

tests/test_model.py
```diff
@@ -514,13 +634,21 @@ def create_update_kwargs(batch_size, optimizer_type, lr, beta1, beta2, epsilon):
 
 
 @pytest.mark.parametrize("n_features", [1, 2])
-def test_bnn_mcmc_update(n_features, hidden_dim_list=(1,), n_samples=5, update_method="MCMC"):
+@pytest.mark.parametrize("dist_type", ["studentt", "normal"])
```
shaharbar1: For mcmc, no need to test normal, the code is fully covered with VI.

tests/test_model.py
```diff
@@ -240,6 +242,79 @@ def test_can_init_studenttarray(shape, mu, sigma, nu):
...
+    """Test that NormalArray inherits from BaseLocationScaleArray."""
+    n = NormalArray.cold_start(shape=5)
+    assert isinstance(n, BaseLocationScaleArray)
+    assert isinstance(n, NormalArray)
+    assert not isinstance(n, StudentTArray)
+
+
+def test_normalarray_to_pymc_distribution():
+    """Test that NormalArray can create PyMC Normal distribution."""
+    import pymc as pm
```
shaharbar1: Move imports to file header

tests/test_model.py
```diff
@@ -240,6 +242,79 @@ def test_can_init_studenttarray(shape, mu, sigma, nu):
     # assert s.shape == shape
 
 
+def test_can_init_normalarray():
```
shaharbar1: These tests are very "cursorish".
Many tests that do same thing, like test_can_init_normalarray and test_can_init_studenttarray. That can be the same test, just with pytest.parametrize

tests/test_model.py
```diff
@@ -240,6 +242,79 @@ def test_can_init_studenttarray(shape, mu, sigma, nu):
...
+    # NormalArray doesn't have nu parameter
+    assert not hasattr(n, "nu")
+
+    # Test with custom args
+    n = NormalArray.cold_start(shape=(3, 4), mu=2.0, sigma=1.5)
+    assert n.mu == np.full((3, 4), 2.0).tolist()
+    assert n.sigma == np.full((3, 4), 1.5).tolist()
+
+
+def test_normalarray_inherits_from_prior_distribution_array():
```
shaharbar1: Redundant test

tests/test_model.py
```diff
@@ -240,6 +242,79 @@ def test_can_init_studenttarray(shape, mu, sigma, nu):
...
+    assert isinstance(n, NormalArray)
+    assert not isinstance(n, StudentTArray)
+
+
+def test_normalarray_to_pymc_distribution():
+    """Test that NormalArray can create PyMC Normal distribution."""
+    import pymc as pm
+    from pymc.distributions.continuous import Normal as PymcNormal
+
+    n = NormalArray.cold_start(shape=(2, 3), mu=0.0, sigma=1.0)
```
shaharbar1: Put this in fixture, you use it multiple times

tests/test_model.py
```diff
@@ -240,6 +242,79 @@ def test_can_init_studenttarray(shape, mu, sigma, nu):
...
+def test_normalarray_sample_rvs():
+    """Test that NormalArray can sample random variates."""
+    n = NormalArray.cold_start(shape=(2, 3), mu=0.0, sigma=1.0)
+    samples = n.sample_rvs(size=(10, 2, 3))
+    assert samples.shape == (10, 2, 3)
+    # Check that samples are reasonable (within a few standard deviations)
+    assert np.all(np.abs(samples) < 10)  # Very unlikely to be outside 10 sigma
+
+
+def test_studenttarray_inherits_from_prior_distribution_array():
```
shaharbar1: Redundant test

tests/test_model.py
shaharbar1: In addition, the tests include many hardcoded consytants instead of utilizing hypothesis random parameter generation.

tests/test_model.py
```diff
@@ -240,6 +242,79 @@ def test_can_init_studenttarray(shape, mu, sigma, nu):
...
+        assert isinstance(pymc_dist.owner.op, PymcStudentT)
+
+
+def test_studenttarray_sample_rvs():
+    """Test that StudentTArray can sample random variates."""
+    s = StudentTArray.cold_start(shape=(2, 3), mu=0.0, sigma=1.0, nu=5.0)
+    samples = s.sample_rvs(size=(10, 2, 3))
+    assert samples.shape == (10, 2, 3)
+    # Check that samples are reasonable
+    assert np.all(np.abs(samples) < 20)  # Student-t has heavier tails
```
shaharbar1: I don't like this. Relies too much on luck. No need to test how np/scipy random number generation works.

shaharbar1 (changes requested):

pybandits/model.py
shaharbar1: Also, you don't need to convert to list all the parameters...

AnastasiiaKabeshova: When I replace it by :
```suggestion
        updated_weight = original_weight.with_dist_parameters(mu=w_mu.tolist(), sigma=w_sigma.tolist())
        updated_bias = original_bias.with_dist_parameters(mu=b_mu.tolist(), sigma=b_sigma.tolist())
```
update test fails

shaharbar1 (approved):

ronshiff1 (approved):

PR #116, opened 2026-01-11
Enhance Bayesian Neural Network with Layerwise Scaling Support and Early Stopping

shaharbar1:
### Changes:
* Added `use_layerwise_scaling` parameter to `StudentTArray` and `BaseBayesianNeuralNetwork` for improved weight initialization.
* Added support for early stopping via update kwargs.
* Updated relevant methods to incorporate layerwise scaling logic.
* Enhanced tests to validate layerwise scaling functionality in Bayesian Neural Networks.

### DIFF
--- a/pybandits/model.py
+++ b/pybandits/model.py
@@ -33,6 +33,7 @@ from pymc import Bernoulli, Data, Deterministic, Minibatch, fit, math, sample
 from pymc import Model as PymcModel
 from pymc import Normal as PymcNormal
 from pymc import StudentT as PymcStudentT
+from pymc.variational.callbacks import CheckParametersConvergence
 from scipy.special import erf
 from scipy.stats import norm, t
 from typing_extensions import Self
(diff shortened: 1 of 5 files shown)


pybandits/model.py
```diff
@@ -633,10 +587,15 @@ def validate_activation(cls, v):
     def approx_history(self) -> Optional[np.ndarray]:
         return self._approx_history
 
-    @property
-    def optimizer(self) -> Callable:
+    def _get_obj_optimizer(self) -> Optional[Callable]:
```
ronshiff1, line 974: why did you change it from property?

pybandits/model.py
```diff
@@ -645,6 +604,16 @@ def optimizer(self) -> Callable:
 
         return _optimizer
 
+    def _get_early_stopping_callback(self) -> Optional[CheckParametersConvergence]:
+        if self.update_kwargs is None:
+            return None
+        early_stopping_kwargs = self.update_kwargs.get("early_stopping_kwargs", None)
+        if early_stopping_kwargs is not None:
+            early_stopping_callback = CheckParametersConvergence(**early_stopping_kwargs)
```
ronshiff1: do we want to check if this succeeds?
also valid for optimizer

pybandits/model.py
```diff
@@ -366,15 +367,32 @@ def cold_start(
...
+        shape: Union[PositiveInt, Tuple[PositiveInt, ...]]
+            Dimensions of the Student's t-distribution array.
+        mu: float
+            Mean of the Student's t-distribution.
+        sigma: NonNegativeFloat
+            Standard deviation of the Student's t-distribution.
+        nu: PositiveFloat
+            Degrees of freedom of the Student's t-distribution.
+        use_layerwise_scaling: bool
+            Whether to use layerwise scaling in the network (default is False).
```
ronshiff1, line 568: you are using a specific method - you need to explain it here

ronshiff1: also, as we discussed, I would set it using cold start (if we want other init method)

pybandits/model.py
```diff
@@ -775,8 +782,12 @@ def create_update_model(
             if batch_size is None:
                 bnn_output = Data("bnn_output", y)
                 bnn_input = Data("bnn_input", x)
+                kwargs = {}
             else:
                 bnn_input, bnn_output = Minibatch(x, y, batch_size=batch_size)
+                # total_size is the total number of samples in the dataset,
+                # including the current batch size and the number of previously seen samples in the dataset
+                kwargs = dict(total_size=x.shape[0] + self.count - 2)
```
ronshiff1: 1. instead of kwargs - I suggest to call it in an informative name - possibly "total_size" or something like that
2. I don't think it's correct to add the self.count - let's discuss this ( the information of the past is embedded inside the priors)

pybandits/model.py
```diff
@@ -917,10 +928,7 @@ def _update(self, context: np.ndarray, rewards: List[BinaryReward]):
                 # variational inference
                 update_kwargs = self.update_kwargs.copy()
```
ronshiff1: shouldn't it be _update_kwargs?

pybandits/model.py
```diff
@@ -1000,6 +1009,8 @@ def cold_start(
             The activation function to use for hidden layers. Supported values are: "tanh", "relu", "sigmoid", "gelu" (default is "tanh").
         use_residual_connections : bool
             Whether to use residual connections in the network (default is False).
+        use_layerwise_scaling : bool
```
ronshiff1, line 1435: again - explain what layerwise scaling is

ronshiff1 (changes requested):

pybandits/model.py
shaharbar1: It is in the cold start. I don't understand

pybandits/model.py
shaharbar1: I wanted to define it in the init, rather then recreate on every update step.

pybandits/model.py
shaharbar1: It is pymc functionality. I don't think I should add anything here. WDYT?

pybandits/model.py
shaharbar1: Correct! Thx!

pybandits/model.py
ronshiff1: let's say we gave wrong early_stopping_kwargs parameters - do we want td try ... except?
also for optimizer...
I think we already did get error because we sent wrong params (I'm not sure how obvious it was)

pybandits/model.py
ronshiff1: I mean, set distribution params directly, and not normalized automatically....
but - do as you want...
Do update the documentation

ronshiff1 (approved):

PR #120, opened 2026-02-15
Add transfer learning functionality

shaharbar1:
### Changes:
* Added transfer learning functionality under pybandits/transfer.py including:
 - Merging two MABs by uniting their actions.
 - Update action model parameters, e.g. for BNN LR change.
 - Edit MAB on the fly including the two above.
 - Expending cmab input features dimension
* Added transfer learning test suite under tests/test_transfer.py

### DIFF
--- a/pybandits/transfer.py
+++ b/pybandits/transfer.py
@@ -1,0 +1,629 @@
+# MIT License
+#
+# Copyright (c) 2023 Playtika Ltd.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""Transfer learning functionality for Multi-Armed Bandits.
+
+This module provides the edit_model_on_the_fly() function for transfer learning,
+which enables updating MAB configuration while preserving learned state and
+automatically handling input dimension changes for contextual MABs.
+"""
+
+import json
+from typing import Any, Dict, Tuple, Type
+
+from loguru import logger
+
+from pybandits.base_model import BaseModelMO, BaseModelSO
+from pybandits.cmab import BaseCmabBernoulli
+from pybandits.mab import BaseMab
+from pybandits.model import BaseBayesianNeuralNetwork
+from pybandits.pydantic_version_compatibility import validate_call
+from pybandits.quantitative_model import QuantitativeModel
+
+
+def _pick_learned_state(source_state: Dict[str, Any], model_cls: Type[BaseModelSO]) -> Dict[str, Any]:
+    """Extract transferable learned parameters from a source state dict.
+
+    Parameters
+    ----------
+    source_state : Dict[str, Any]
+        Source action state dictionary.
+    model_cls : Type[BaseModelSO]
+        Model class whose transfer_learned_keys declares which keys to extract.
+
+    Returns
+    -------
+    Dict[str, Any]
+        Subset of source_state containing only the declared learned keys.
+    """
+    return {k: source_state[k] for k in model_cls.transfer_learned_keys if k in source_state}
+
+
+def _get_state_dict(mab: BaseMab) -> Dict[str, Any]:
+    """Convert MAB to dictionary via JSON serialization.
+
+    Parameters
+    ----------
+    mab : BaseMab
+        MAB instance to convert.
+
+    Returns
+    -------
+    state_dict : Dict[str, Any]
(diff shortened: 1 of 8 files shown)


pybandits/transfer.py
```diff
@@ -0,0 +1,757 @@
...
+_RESTRICTED_KEYS = frozenset(
+    [
+        "n_successes",  # Learned state
+        "n_failures",  # Learned state
+        "model_params",  # Learned weights
+        "bnn_layer_params",  # Learned weights (alias)
+        "bnn_layer_params_init",  # Frozen initial state
+        "dimension",  # Structural
+        "activation",  # Structural
+        "use_residual_connections",  # Structural
```
fsimond: You need to add "cost" here

pybandits/transfer.py
```diff
@@ -0,0 +1,757 @@
...
+    return set(actions1.keys()) & set(actions2.keys())
+
+
+def _validate_mab_compatibility(mab1: BaseMab, mab2: BaseMab) -> None:
+    """Validate that two MABs can be merged.
+
+    Checks that both MABs have:
+    - Same class type
+    - Same ActionsManager type (validated by Pydantic during reconstruction)
+    - Compatible model configurations
```
fsimond: You don't check "compatible model configurations" in this function

pybandits/transfer.py
```diff
@@ -0,0 +1,757 @@
...
+    if mab1_manager_type != mab2_manager_type:
+        raise ValueError(
+            f"Cannot merge MABs with different ActionsManager types: "
+            f"{mab1_manager_type} and {mab2_manager_type}. "
+            "Both MABs must have the same variant (SO/MO/CC/MOCC)."
+        )
+
+
+def _deep_merge_dicts(base: Dict[str, Any], updates: Dict[str, Any]) -> Dict[str, Any]:
+    """Deep merge updates into base dictionary.
```
fsimond: potential risk in this function.
`base = {
    "params": {
        "a": 1,
        "b": 2,
        "c": 3
    },
}`

`updates = {
    "params": {
        "a": 42
    },
}
`

Will give you 
`{
    "params": {
        "a": 42,
        "b": 2,
        "c": 3
    }
}
`


Don't you want to also remove the missing params ? "b" and "c" ?

That's what we were doing to update params previously

pybandits/transfer.py
```diff
@@ -0,0 +1,757 @@
...
+    _validate_mab_compatibility(mab1, mab2)
+    state1 = _get_state_dict(mab1)
+    state2 = _get_state_dict(mab2)
+
+    # Extract actions
+    actions1 = state1["actions_manager"]["actions"]
+    actions2 = state2["actions_manager"]["actions"]
+
+    # Check for conflicts
+    conflicts = set(actions1.keys()) & set(actions2.keys())
```
fsimond: I think we **want** to support the merge of models with the same action_ids.
This is actually the exact purpose, we take the same model, only built with different params. 

But they will have the same action ids

pybandits/transfer.py
```diff
@@ -0,0 +1,861 @@
...
+from pybandits.pydantic_version_compatibility import validate_call
+
+# Parameters that are safe to edit (hyperparameters and configuration)
+# All other parameters are considered non-editable (learned state or structural)
+_EDITABLE_KEYS = frozenset(
+    [
+        "epsilon",  # MAB-level exploration rate
+        "delta",  # ActionsManager-level adaptive windowing parameter
+        "update_kwargs",  # Action-level training configuration (learning rate, epochs, etc.)
+        "update_method",  # Action-level inference method (VI vs MCMC)
```
fsimond: there are lots of other parameters that I thought we could modify:
subsidy_factor (I guess this one is missing for sure)

But what about
- hidden_dim_list
- dist_params_init (if you change update_method), you will also need to change dist_params_init no ? as one of the two method does not support "nu" for example
- dist_type
- activation
- use_residual_connections
- use_layerwise_scaling

pybandits/transfer.py
```diff
@@ -0,0 +1,861 @@
...
+    Returns
+    -------
+    mab : BaseMab
+        Reconstructed MAB instance.
+    """
+    state_json = json.dumps(state_dict)
+    return mab_class.from_state(state_json)
+
+
+def _check_action_conflicts(actions1: Dict[ActionId, Any], actions2: Dict[ActionId, Any]) -> Set[ActionId]:
```
fsimond: can be removed now

pybandits/transfer.py
```diff
@@ -0,0 +1,861 @@
...
+            merged_actions[action_id] = action_state
+
+    logger.info(
+        f"Merged {type(mab1).__name__}: used mab2 as template with {len(actions2)} action(s), "
+        f"transferred learned state from mab1 for {len(overlapping)} overlapping action(s)."
+    )
+
+    # Create merged state (use strategy/epsilon from mab1)
+    # We can directly mutate state1 since it's from JSON and won't be reused
+    merged_state = state1
```
fsimond: I think you should use state2 here, to preserve the model-level configuration of Mab2

pybandits/transfer.py
```diff
@@ -0,0 +1,861 @@
...
+    # Merge actions_with_change sets (for adaptive windowing)
+    actions_with_change_1 = set(state1["actions_manager"].get("actions_with_change", []))
+    actions_with_change_2 = set(state2["actions_manager"].get("actions_with_change", []))
+    merged_state["actions_manager"]["actions_with_change"] = list(actions_with_change_1 | actions_with_change_2)
+
+    # Reconstruct MAB
+    return _reconstruct_mab(type(mab1), merged_state)
+
+
+def _extract_action_state(mab: BaseMab) -> Dict[str, Any]:
```
fsimond: unused, can be removed

pybandits/transfer.py
```diff
@@ -0,0 +1,861 @@
...
+    filtered_state = {key: value for key, value in first_action_state.items() if key in _EDITABLE_KEYS}
+
+    logger.debug(
+        f"Extracted {len(filtered_state)} editable parameters from action model (out of {len(_EDITABLE_KEYS)} possible)"
+    )
+    return filtered_state
+
+
+@validate_call
+def update_action_models_parameters(
```
fsimond: This will indeed not be used at the moment, you can decide if you want to remove it

pybandits/transfer.py
```diff
@@ -0,0 +1,861 @@
...
+        logger.info(
+            f"Updated {len(action_params)} parameter(s) for all {len(actions)} actions: {list(action_params.keys())}"
+        )
+
+    # Reconstruct MAB
+    return _reconstruct_mab(type(mab), state_dict)
+
+
+@validate_call
+def edit_model_on_the_fly(
```
fsimond, line 488: indeed, this function is just a wrapper for merge_mabs, you can decide which one you wanna keep

fsimond (approved):
LGTM :)

PR #122, opened 2026-02-27
Add Categorical Embeddings to BNN

AnastasiiaKabeshova:
### changes:
* Update pybandits/model.py
  - Add CategoricalFeatureConfig (column_index, cardinality, embedding_dim)
    and FeaturesConfig for describing mixed numerical/categorical context
    arrays with validation
  - Add EmbeddingParams: Bayesian embedding matrices per categorical
    feature, jointly trained via VI/MCMC with cold_start and reset
  - Refactor BaseLocationScaleArray: add public param_map, _sampler,
    _to_sampler_kwargs, sample_at_indices; StudentTArray.param_map
    inherits via {**BaseLocationScaleArray.param_map, "nu": "df"}
  - Refactor BaseBayesianNeuralNetwork to accept FeaturesConfig; add
    sample_embeddings, extract_sample, _prepare_forward_input,
    _prepare_bnn_input_output, _apply_bnn_layers,
    _make_embedding_matrix_inputs
  - Separate sampling from forward pass: sample_proba calls
    sample_weights + sample_embeddings, then deterministic forward_pass
  - create_model_params takes dist_class (type[BaseLocationScaleArray])
    instead of dist_type literal
  - cold_start accepts categorical_features dict; embedding_dim
    auto-computed via _embedding_dim_divisor class var
  - Add backward-compatible state migration and reset for embedding
    params and feature_config

* Update pybandits/cmab.py
  - Import BaseLocationScaleArray; use param_map instead of hardcoded
    "mu" key in update_old_state
  - Add feature_config migration for 3.0-5.0 model states

* Update pybandits/quantitative_model.py
  - Update QBNN cold_start to accept categorical_features
    (Dict[NonNegativeInt, NonNegativeInt]) with auto column-index
    shifting by dimension
  - Pre-sample weights and embeddings in sample_proba so closures are
    deterministic across quantity evaluations
  - Simplify _to_quantitative_probabilities using bnn.extract_sample

* Update tests/test_model.py
  - Add tests for CategoricalFeatureConfig, FeaturesConfig validation,
    EmbeddingParams cold_start shapes, embedding init frozen copy,
    categorical context validation, and BNN with embeddings end-to-end
  - Simplify FeaturesConfig construction (drop redundant
    categorical_features_configs=[])

* Update tests/test_quantitative_model.py
  - Add tests for QBNN with categorical features: initialization,
    sample_proba, update, and closure determinism

* Update docs/src/tutorials/bnn.ipynb
  - Add FeaturesConfig / categorical embeddings tutorial section with
    embedding inspection and uncertainty maps

**Summary**
This PR introduces Bayesian categorical embeddings to the BNN model, enabling contextual bandits to consume mixed-type input arrays where some columns are integer-encoded categorical features. Categorical columns are modelled with learnable Bayesian embedding matrices inside the PyMC graph — the same StudentTArray/NormalArray posterior representation used for all other BNN parameters.

This is a breaking change in how BayesianNeuralNetwork is constructed and serialised. The version is bumped to 5.0.0.

**How the embedding is wired into the BNN**
```
Input array  (n_samples, n_features)
      │
      ├─► numerical columns ──────────────────────────────────────────────┐
      │                                                                     │
      └─► categorical column k ──► EmbeddingMatrix[k] (vocab_size, emb_dim) ─► lookup ─► (n_samples, emb_dim) ─┘
                                                                                                                 │
                                                                              np.concatenate ◄──────────────────┘
                                                                                     │
                                                                            (n_samples, n_numerical + Σ emb_dim)
                                                                                     │
                                                                              BNN layers (unchanged)

```

**Usage**
```
from pybandits.model import (
    BayesianNeuralNetwork, FeatureConfig, CategoricalFeatureConfig
)

# Input layout: column 0 = score (float), column 1 = tier (int, 3 categories)
fc = FeatureConfig(
    n_features=2,
    categorical_features_configs=[
        CategoricalFeatureConfig(column_index=1, vocab_size=3, embedding_dim=8),
    ],
)

bnn = BayesianNeuralNetwork.cold_start(feature_config=fc, hidden_dim_list=[16])

context = np.array([[0.5, 0], [-0.3, 2], [1.2, 1]])   # tier index in column 1
results = bnn.sample_proba(context)
```

**Tests**
31 new tests covering CategoricalFeatureConfig, FeatureConfig, EmbeddingParams, sample_proba, create_update_model, _update, _reset, from_old_state

## Summary by CodeRabbit

* **New Features**
  * Added support for categorical feature embeddings in Bayesian Neural Networks
  * Enabled mixed numerical and categorical input handling
  * Added optional title parameter to visualization functions

* **Documentation**
  * Enhanced tutorial with comprehensive Bayesian Neural Network workflows featuring categorical embeddings and mixed feature types

* **Chores**
  * Version bump to 5.0.0


docs/src/tutorials/bnn.ipynb
```diff
@@ -2,39 +2,28 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "id": "0",
```
shaharbar1, line 5: Why do we have all these "id" fields now?

pybandits/model.py
```diff
@@ -797,6 +797,124 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+    The context array is expected to have columns in this fixed order:
+    ``[numerical_0, ..., numerical_{n-1}, cat_0, cat_1, ...]``
+
+    - The first ``n_numerical`` columns are numerical (pass-through, float).
+    - The following ``len(categorical_features)`` columns are integer-encoded
+      categorical indices (pre-encoded by the caller, 0-based).
+
+    Parameters
+    ----------
+    n_numerical : int
```
shaharbar1: It is problematic that numerical features are assumed to be in the vector beginning and categorical features at the end. For example, it conflicts with what happens on transfer learning scenario. You should consult with @fsimond and agree on the right way to do it.

pybandits/model.py
```diff
@@ -797,6 +797,124 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+        Number of leading numerical columns. Default 0.
+    categorical_features : List[CategoricalFeatureConfig]
+        Ordered list of categorical feature configurations. Each entry corresponds
+        to one column (in order) after the numerical columns.
+    """
+
+    n_numerical: int = 0
```
shaharbar1: n_numerical_features

pybandits/model.py
```diff
@@ -797,6 +797,124 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+    categorical_features: List[CategoricalFeatureConfig] = Field(default_factory=list)
```
shaharbar1: categorical_features_configs

pybandits/model.py
```diff
@@ -797,6 +797,124 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+
+    @property
+    def total_input_dim(self) -> int:
```
shaharbar1: No need for the "total" prefix IMO

pybandits/model.py
```diff
@@ -903,6 +1030,7 @@ class BaseBayesianNeuralNetwork(Model, ABC):
     _prob_var_name: ClassVar[str] = "prob"
     weight_var_name: ClassVar[str] = "weight"
     _bias_var_name: ClassVar[str] = "bias"
+    _emb_var_name: ClassVar[str] = "emb"
```
shaharbar1: I prefer the name "embedding", I don't think this shortcut is clear enough.

pybandits/model.py
```diff
@@ -938,6 +1066,7 @@ class BaseBayesianNeuralNetwork(Model, ABC):
     update_kwargs: Optional[dict] = None
     activation: ActivationFunctions = "tanh"
     use_residual_connections: bool = False
+    feature_config: Optional[FeatureConfig] = None
```
shaharbar1: Is it possible that a model will get None value for feature_config? It doesn't makes sense

pybandits/model.py
```diff
@@ -975,10 +1104,51 @@ def validate_activation(cls, v):
...
+            Float array of numerical features. Empty ``(n_samples, 0)`` if none.
+        cat_indices_dict : Dict[int, np.ndarray]
+            Mapping from feature index (0-based) to int32 array of shape ``(n_samples,)``
+            containing integer category codes.
+        """
+        n_samples = len(context)
+        n_numerical = self.feature_config.n_numerical
+
+        if n_numerical > 0:
+            numerical_arr = context[:, :n_numerical].astype(np.float64)
```
shaharbar1: Why float64? What dtype did we used so far?

pybandits/model.py
```diff
@@ -975,10 +1104,51 @@ def validate_activation(cls, v):
...
+        else:
+            numerical_arr = np.empty((n_samples, 0), dtype=np.float64)
+
+        cat_indices_dict: Dict[int, np.ndarray] = {}
+        for i in range(len(self.feature_config.categorical_features)):
```
shaharbar1: I don't need you really need a loop in here. You can make this code smoother and faster with dict(zip())

pybandits/model.py
```diff
@@ -1035,15 +1206,20 @@ def create_model_params(
 
         Parameters
         ----------
-        n_features : PositiveInt
-            The number of input features for the BNN.
+        n_features : Optional[PositiveInt]
```
shaharbar1: In our usage this is an internal function. You can avoid having n_features. Have the outside function define feature_config on the fly.

pybandits/model.py
```diff
@@ -1070,33 +1247,55 @@ def create_model_params(
...
             raise ValueError("Context array must contain only numeric values.")
 
-        if n_cols_context != self.input_dim:
-            raise AttributeError(f"Shape mismatch: context must have {self.input_dim} columns.")
+        if n_cols_context != expected_cols:
+            raise AttributeError(f"Shape mismatch: context must have {expected_cols} columns.")
+
+        if self.feature_config is not None:
+            n_num = self.feature_config.n_numerical
+            for i, cat_cfg in enumerate(self.feature_config.categorical_features):
```
shaharbar1: Again I don't think you need a loop here. As for the ValueError, it can be done with numpy

pybandits/model.py
```diff
@@ -1157,127 +1356,194 @@ def model_post_init(self, __context: Any) -> None:
...
+        When ``feature_config`` is set, ``x`` must be a numpy array where the first
+        ``n_numerical`` columns are numerical features and subsequent columns are integer
+        category indices. Categorical features are modelled with Bayesian embedding matrices
+        inside the PyMC graph.
+
         Parameters
         ----------
         x : ArrayLike
-            Input features of shape (n_samples, n_features)
+            Input features. Shape ``(n_samples, n_features)``.
```
shaharbar1: Why?

pybandits/model.py
```diff
@@ -1157,127 +1356,194 @@ def model_post_init(self, __context: Any) -> None:
...
-        Notes
-        -----
-        The model structure follows these steps:
-        1. For each layer, create weight and bias variables from prior distributions.
-        2. Apply linear transformations and activations through the layers.
-        3. Apply sigmoid activation at the output
-        4. Use Bernoulli likelihood for binary classification
+            PyMC model object with the specified neural network architecture.
         """
+        has_categorical = self.feature_config is not None and bool(self.feature_config.categorical_features)
```
shaharbar1: Should be a propery of FeatureConfig I think

pybandits/model.py
```diff
@@ -1157,127 +1356,194 @@ def model_post_init(self, __context: Any) -> None:
...
+        if batch_size is not None and has_categorical:
+            raise NotImplementedError(
+                "Minibatch training with categorical embedding features is not yet supported. "
+                "Set batch_size=None or remove categorical features from feature_config."
+            )
+
         y = np.array(y, dtype=np.int32)
+
+        if self.feature_config is None:
+            # --- Legacy path (unchanged behaviour) ---
```
shaharbar1: No. Don't do that. That's just spamming our code base. No legacy path like this. There's no reason here. If you defined feature_config use it always, Not part of the time. If you think it is a big change, alter pybandits version to 5.0.0 (I don't think that's the case, but I'm not sure).

pybandits/model.py
```diff
@@ -1157,127 +1356,194 @@ def model_post_init(self, __context: Any) -> None:
...
+            bnn_output = Data("bnn_output", y)
+
+            parts = []
+
+            # Numerical features
+            if numerical_arr.shape[1] > 0:
+                parts.append(Data("bnn_numerical_input", numerical_arr))
+
+            # Categorical Bayesian embeddings
+            for i, cat_cfg in enumerate(self.feature_config.categorical_features):
```
shaharbar1: Put the inner loop code into a function and use list comprehension to build parts.

pybandits/model.py
```diff
@@ -1157,127 +1356,194 @@ def model_post_init(self, __context: Any) -> None:
...
+                emb_var_name = self.get_embedding_var_name(i)
+                emb_dist = self.model_params.embedding_params.embeddings[i]
+                emb_shape = emb_dist.shape  # (vocab_size, emb_dim)
+
+                emb_matrix = emb_dist.to_pymc_distribution(name=emb_var_name, shape=emb_shape)
+                idx_data = Data(f"cat_idx_{i}", cat_indices_dict[i])
+                parts.append(emb_matrix[idx_data])  # (n_samples, emb_dim)
```
shaharbar1: parts is a bad name.

pybandits/model.py
```diff
@@ -1157,127 +1356,194 @@ def model_post_init(self, __context: Any) -> None:
...
-                            next_layer_input = activated_output + residual_padded
-                    else:
-                        next_layer_input = activated_output
+                next_layer_input = math.concatenate(parts, axis=1)
 
-            # Final output processing
+            linear_transform, _ = self._build_bnn_layers(next_layer_input)
             logit = Deterministic(self._logit_var_name, linear_transform.squeeze())
-
-            Bernoulli("out", logit_p=logit, observed=bnn_output, **likelihood_kwargs)
```
shaharbar1, line 1237: Bad change. You will break KLD weighing this way.

pybandits/model.py
```diff
@@ -1157,127 +1356,194 @@ def model_post_init(self, __context: Any) -> None:
...
         """
-
-        # check input args
         self.check_context_matrix(context=context)
 
-        _context = np.atleast_2d(context)
-        n_samples = len(_context)
-        # Sample from distributions for each layer
-        next_layer_input = _context
+        if self.feature_config is None:
```
shaharbar1: I don't like this at all. Make the code more sustainable.

shaharbar1 (changes requested):
With great power (using cursor/claude code) comes great responsibility (for keeping the coding standards of this repo).

pybandits/model.py
AnastasiiaKabeshova: we talked with Florian, and agreed that the best solution would be to pass indexes of the categorical features on the initialisation stage

pybandits/model.py
AnastasiiaKabeshova: Yes. A model can have feature_config is None. You can create a BNN with n_features=5 and feature_config=None (numerical-only).

So Optional[FeatureConfig] = None is correct: None means “no feature config / numerical-only BNN.” No change is required for that field.

pybandits/model.py
AnastasiiaKabeshova: just fixing a formulation nothing special :)
the purpose was to add missing batch_size param in the comment

Periecle:
@coderabbitai review

pybandits/model.py
shaharbar1: I think that's ambigious and bad practice. Model init should have only feature_config, while cold start should get number of numerical features and categorical features config. Cold start should then build features_config and serve it to init.

pybandits/model.py
AnastasiiaKabeshova: changed

pybandits/model.py
AnastasiiaKabeshova: I wanted to keep feature_config  optional, if I make it not optional we need to increase the major version of library.
what is the issue you have with this one?

pybandits/model.py
```diff
@@ -796,6 +825,155 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+    Parameters
+    ----------
+    n_features : int
+        Total number of columns in the input numpy array. Default 0.
+    categorical_features_configs : List[CategoricalFeatureConfig]
+        List of categorical feature configurations. Each entry carries a
+        ``column_index`` indicating its position in the input array.
+    """
+
+    n_features: int = 0
```
ronshiff1: n_features should be either positive or non negative

pybandits/model.py
```diff
@@ -796,6 +825,160 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+    embedding_dim : PositiveInt
+        Dimensionality of the embedding vector for this feature. Default is 8.
+    """
+
+    column_index: NonNegativeInt
+    vocab_size: PositiveInt
+    embedding_dim: PositiveInt = 8
+
+
+class FeatureConfig(PyBanditsBaseModel):
```
ronshiff1: I think FeautresConfig is a better name - becuase it manages the config of all features and not a specific one

pybandits/model.py
```diff
@@ -975,10 +1169,53 @@ def validate_activation(cls, v):
...
+        n_samples = len(context)
+        num_indices = self.feature_config.numerical_indices
+
+        if num_indices:
+            numerical_arr = context[:, num_indices]
+        else:
+            numerical_arr = np.empty((n_samples, 0), dtype=context.dtype)
+
+        configs = self.feature_config.categorical_features_configs
+        cat_indices_dict = dict(
```
ronshiff1: I think it can be done directly with 
cat_indices_dict = {i : context[:, cfg.column_index].astype(np.int32) for i,cfg in enumerate(configs)}
to me it's a little bit mode readable. but whatever you like

docs/src/tutorials/bnn.ipynb
AnastasiiaKabeshova: I do not know :)
pre-commit passed, this was not removed

pybandits/model.py
```diff
@@ -1594,6 +1979,49 @@ class BayesianNeuralNetwork(BaseBayesianNeuralNetwork):
...
+
+        _context = np.atleast_2d(context)
+        numerical_arr, cat_indices_dict = self._prepare_context_arrays(_context)
+
+        parts = [numerical_arr] if numerical_arr.shape[1] > 0 else []
+        for i in range(len(self.feature_config.categorical_features_configs)):
+            emb_dist = self.model_params.embedding_params.embeddings[i]
+            indices = cat_indices_dict[i]
+            # reshape to (n_samples, emb_dim) — scipy.rvs squeezes the batch dim when n_samples=1
+            emb_vectors = emb_dist.sample_at_indices(indices).reshape(len(indices), -1)
```
ronshiff1: This will break the QBNN algorithm - 
since in case of categorical features,  every forward pass will yield a different result.
we need to make the forward pass a function only of quantities in the QBNN algorithm

ronshiff1 (changes requested):

tests/test_quantitative_model.py
ronshiff1, line 437: please make this also for the categorical features case

docs/src/tutorials/bnn.ipynb
shaharbar1: Change nbstripout pre commit hook in the repo to:
- id: nbstripout
  args: [
    '--extra-keys',
    'cell.id metadata.kernelspec metadata.language_info.version metadata.language_info.pygments_lexer'
  ]

pybandits/cmab.py
```diff
@@ -263,9 +264,23 @@ def update_old_state(
                 )._apply_version_adjusted_method("model_dump", "dict")
                 action_state["model_params"] = model_params
 
+                feature_config = FeatureConfig(
+                    n_features=len(mu_list), categorical_features_configs=[]
+                )._apply_version_adjusted_method("model_dump", "dict")
```
shaharbar1: Make _apply_version_adjusted_method unprotected by removing the underscore prefix

pybandits/cmab.py
```diff
@@ -263,9 +264,23 @@ def update_old_state(
...
+                action_state["feature_config"] = feature_config
+
                 action_state.pop("alpha")
                 action_state.pop("betas")
 
+            # 3.0 <= version < 4.4: model_params present but feature_config absent (numerical-only).
```
shaharbar1: < 5, not < 4.4

pybandits/cmab.py
```diff
@@ -263,9 +264,23 @@ def update_old_state(
...
+            elif "model_params" in action_state and "feature_config" not in action_state:
+                layer_params = action_state["model_params"]["bnn_layer_params"]
+                n_features = len(layer_params[0]["weight"]["mu"])
```
shaharbar1: Use BaseBayesianNeuralNetwork.weight_var_name instead of "weight"

pybandits/model.py
```diff
@@ -1418,8 +1717,34 @@ def _create_updated_layer_params(
...
+            emb_var_name = self.get_embedding_var_name(i)
+            emb_values = trace[emb_var_name]
+            emb_mu = np.mean(emb_values, axis=0)
+            emb_sigma = np.std(emb_values, axis=0)
+            updated_embeddings.append(orig_emb.with_dist_parameters(mu=emb_mu.tolist(), sigma=emb_sigma.tolist()))
+        self.model_params.embedding_params.embeddings = updated_embeddings
+
     @validate_call(config=dict(arbitrary_types_allowed=True))
-    def _update(self, context: np.ndarray, rewards: List[BinaryReward]):
+    def _update(self, context: Any, rewards: List[BinaryReward]):
```
shaharbar1: Why did you change to Any?

pybandits/model.py
```diff
@@ -725,6 +746,13 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
             size=size,
         )
 
+    def sample_at_indices(self, indices: np.ndarray) -> np.ndarray:
+        return t.rvs(
```
shaharbar1: I think some refactoring is needed here.
You use t.rvs / norm.rvs multiple times. It should be defined via classvar as "sampler" or something like that.
Then, sample_at_indices & sample_rvs should not be abstract at BaseLocationScaleArray, they should call self._params to make the sampling with the required parameter (size / indices)

pybandits/model.py
```diff
@@ -797,6 +825,155 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+    vocab_size : PositiveInt
+        Number of distinct integer category codes. The context array must contain
+        pre-encoded integer indices in the range ``[0, vocab_size)``.
+    embedding_dim : PositiveInt
+        Dimensionality of the embedding vector for this feature. Default is 8.
+    """
+
+    column_index: NonNegativeInt
+    vocab_size: PositiveInt
+    embedding_dim: PositiveInt = 8
```
shaharbar1: Why 8???

pybandits/model.py
```diff
@@ -797,6 +825,155 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+        Column position of this feature in the input numpy array.
+    vocab_size : PositiveInt
+        Number of distinct integer category codes. The context array must contain
+        pre-encoded integer indices in the range ``[0, vocab_size)``.
+    embedding_dim : PositiveInt
+        Dimensionality of the embedding vector for this feature. Default is 8.
+    """
+
+    column_index: NonNegativeInt
+    vocab_size: PositiveInt
```
shaharbar1: replace with "cardinality". Vocabulary is more suitable for NLP.

pybandits/model.py
```diff
@@ -797,6 +825,155 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+        if isinstance(values, dict):
+            if not values.get("embeddings_init"):
+                values["embeddings_init"] = deepcopy(values.get("embeddings", []))
+        return values
+
+    @classmethod
+    def cold_start(
+        cls,
+        feature_config: "FeatureConfig",
+        dist_class: type = StudentTArray,
```
shaharbar1: Use dist_type literal like in the BNN cold start

pybandits/model.py
```diff
@@ -797,6 +825,155 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+            Feature configuration containing categorical feature specs.
+        dist_class : type
+            Distribution class to use (``StudentTArray`` or ``NormalArray``).
+        **dist_params_init
+            Distribution parameters passed to ``BaseLocationScaleArray.cold_start``
+            (e.g. ``mu``, ``sigma``, ``nu`` for StudentT; ``mu``, ``sigma`` for Normal).
+
+        Returns
+        -------
+        EmbeddingParams
```
shaharbar1, line 923: Add some detail

pybandits/model.py
```diff
@@ -976,10 +1164,53 @@ def validate_activation(cls, v):
...
+        Returns
+        -------
+        numerical_arr : np.ndarray of shape (n_samples, n_numerical)
+            Numerical features, same dtype as input. Empty ``(n_samples, 0)`` if none.
+        cat_indices_dict : Dict[int, np.ndarray]
+            Mapping from feature index (0-based) to int32 array of shape ``(n_samples,)``
+            containing integer category codes.
+        """
+        n_samples = len(context)
+        num_indices = self.feature_config.numerical_indices
```
shaharbar1: num is confusing, use numeric

pybandits/model.py
```diff
@@ -1022,25 +1253,24 @@ def get_layer_params_name(cls, layer_ind: PositiveInt) -> Tuple[str, str]:
     @classmethod
     def create_model_params(
         cls,
-        n_features: PositiveInt,
+        feature_config: Union[FeatureConfig, PositiveInt],
```
shaharbar1: Just make it FeatureConfig. No need to support the old scenario

pybandits/model.py
```diff
@@ -1022,25 +1253,24 @@ def get_layer_params_name(cls, layer_ind: PositiveInt) -> Tuple[str, str]:
...
         hidden_dim_list: Optional[List[PositiveInt]],
         use_layerwise_scaling: bool = False,
         dist_class: type[BaseLocationScaleArray] = StudentTArray,
         **dist_params_init,
     ) -> BnnParams:
         """
-        Creates model parameters for a Bayesian neural network (BNN) model according to dist_params_init
-        This method initializes the distribution's parameters for each layer of a BNN
-        using the specified number of features, hidden dimensions, and distribution
-        initialization parameters.
```
shaharbar1, line 1035: Why did you change the docstring?

pybandits/model.py
```diff
@@ -1022,25 +1253,24 @@ def get_layer_params_name(cls, layer_ind: PositiveInt) -> Tuple[str, str]:
...
+        Creates model parameters for a Bayesian neural network (BNN) model according to dist_params_init.
 
         Parameters
         ----------
-        n_features : PositiveInt
-            The number of input features for the BNN.
+        feature_config : Union[FeatureConfig, PositiveInt]
+            Either a ``FeatureConfig`` describing the full input layout, or a plain integer
+            ``n_features`` for backwards compatibility (treated as an all-numerical config).
```
shaharbar1: No need for backward compatibility. Stick with FeatureConfig.

pybandits/model.py
```diff
@@ -1071,46 +1304,72 @@ def create_model_params(
...
+            )
 
-    @validate_call(config=dict(arbitrary_types_allowed=True))
-    def check_context_matrix(self, context: np.ndarray):
+        return BnnParams(
+            bnn_layer_params=layer_params_init,
+            embedding_params=embedding_params,
+        )
+
+    def check_context_matrix(self, context: Any):
```
shaharbar1: Why did you make it any?
Cmab class should already make it np array.

pybandits/model.py
```diff
@@ -1158,83 +1417,124 @@ def model_post_init(self, __context: Any) -> None:
...
+                    offset += 1
+                cat_mbs = [mb_results[offset + i] for i in range(len(configs))]
+                bnn_output = mb_results[offset + len(configs)]
+
+                for i, cat_cfg in enumerate(configs):
+                    emb_var_name = self.get_embedding_var_name(i)
+                    emb_dist = self.model_params.embedding_params.embeddings[i]
+                    emb_shape = emb_dist.shape  # (vocab_size, emb_dim)
+                    emb_matrix = emb_dist.to_pymc_distribution(name=emb_var_name, shape=emb_shape)
+                    input_parts.append(emb_matrix[cat_mbs[i]])  # (batch_size, emb_dim)
```
shaharbar1: This code snippet should be exported to a designated function with good docstring. In addition the code repeats itself for the case of no minibatches. You can unify it further for the two cases.

pybandits/model.py
```diff
@@ -1418,8 +1717,34 @@ def _create_updated_layer_params(
 
         return BnnLayerParams(weight=updated_weight, bias=updated_bias)
 
+    def _update_embedding_params_from_mapping(self, mu_sigma_mapping: Dict[str, Tuple[np.ndarray, np.ndarray]]) -> None:
+        """Update embedding matrices from a ``{var_name: (mu_array, sigma_array)}`` mapping (VI path)."""
```
shaharbar1: Elaborate the docstring....

pybandits/model.py
```diff
@@ -1514,19 +1843,21 @@ def cold_start(
         activation: ActivationFunctions = "tanh",
         use_residual_connections: bool = False,
         use_layerwise_scaling: bool = False,
+        feature_config: Optional[FeatureConfig] = None,
```
shaharbar1: I don't like leaving both n_features and feature_config.
One of the principals in the design of cold start method is easy interface, with no internal classes (like FeatureConfig). I suggest to have n_features along with a dict representation of the categorical config. Then, cold start would build FeatureConfig internally before serving to the model.
As for the embedding dim. A good rule of time is to make it ceiling(cardinality // 4). You can use it in cold start, and just serve index and cardinality as a dict.

pybandits/model.py
```diff
@@ -1576,14 +1919,56 @@ def cold_start(
...
+        >>> import json
+        >>> raw = json.loads(old_model.get_state())   # pre-4.4 JSON
+        >>> model = BayesianNeuralNetwork.from_old_state(raw)
+        """
+        if "feature_config" not in state:
+            layer_params = state["model_params"]["bnn_layer_params"]
+            if not layer_params:
+                raise ValueError("Cannot infer feature_config: bnn_layer_params is empty.")
+            # weight.mu is List[List[float]]; outer length == input_dim == n_features for numerical-only models
+            n_features = len(layer_params[0]["weight"]["mu"])
```
shaharbar1: replace weight with weight_var_name

pybandits/model.py
```diff
@@ -1576,14 +1919,56 @@ def cold_start(
...
+            fc = FeatureConfig(n_features=n_features, categorical_features_configs=[])
+            state = {**state, "feature_config": fc._apply_version_adjusted_method("model_dump", "dict")}
+
+        if pydantic_version == PYDANTIC_VERSION_2:
```
shaharbar1, line 2045: Check also for pydantic 1 and raise value error like we do in rest of the repo.

shaharbar1 (changes requested):

pybandits/model.py
ronshiff1: Totally agree!

pybandits/model.py
```diff
@@ -1594,6 +1979,49 @@ class BayesianNeuralNetwork(BaseBayesianNeuralNetwork):
     modeling and inference using neural networks.
     """
 
+    def forward_pass(
```
ronshiff1: why isn't it in the BaseBayesianNeuralNetwork class?
(we want all classes which inherit from BaseBayesianNeuralNetwork to have it)

pybandits/model.py
```diff
@@ -1158,83 +1417,124 @@ def model_post_init(self, __context: Any) -> None:
...
+        numerical_arr, cat_indices_dict = self._prepare_context_arrays(x)
+        configs = self.feature_config.categorical_features_configs
+
         with PymcModel() as _model:
-            # Define data variables
-            if batch_size is None:
-                bnn_output = Data("bnn_output", y)
-                bnn_input = Data("bnn_input", x)
-                likelihood_kwargs = {}
+            if batch_size is not None:
```
ronshiff1: We're adding a lot of code here (in every feature we multiply the code by 2).
let's try to make functions small by adding close understandable functions

pybandits/model.py
AnastasiiaKabeshova: done

pybandits/model.py
AnastasiiaKabeshova: why not? is it too big from your point of view?
what would be your suggestion for default value?

pybandits/model.py
AnastasiiaKabeshova: forgot to change it back

pybandits/model.py
AnastasiiaKabeshova: I am agree that cardinality is the more precise term for this statistical context

That said, vocab_size is instantly recognizable to anyone who used Embedding in diff libraries, so it has strong precedent in the embedding literature specifically.

would suggest to leave it like that :)

pybandits/model.py
ronshiff1: like - create_minibatch function
this code repeats itself:
for i, cat_cfg in enumerate(configs):
emb_var_name = self.get_embedding_var_name(i)
emb_dist = self.model_params.embedding_params.embeddings[i]
emb_shape = emb_dist.shape # (vocab_size, emb_dim)
emb_matrix = emb_dist.to_pymc_distribution(name=emb_var_name, shape=emb_shape)
input_parts.append(emb_matrix[cat_mbs[i]]) # (batch_size, emb_dim)
why not wrappiig it in a function?

pybandits/model.py
AnastasiiaKabeshova: I found it too obvious :) 
not very informative

pybandits/quantitative_model.py
ronshiff1: This is something I actaully did.
can you remove the "Union"? since it's one type...
it repeats several times

pybandits/model.py
```diff
@@ -369,6 +369,27 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
         """
         pass
 
+    @abstractmethod
+    def sample_at_indices(self, indices: np.ndarray) -> np.ndarray:
```
ronshiff1: I think it's better to define indices as List[NonNegativeInt]

pybandits/model.py
AnastasiiaKabeshova: yes, It is a leftover from previous solution 
will change

pybandits/model.py
AnastasiiaKabeshova: ok, removed feature_config from here

AnastasiiaKabeshova: also added categorical_features

AnastasiiaKabeshova: **Before**
BayesianNeuralNetwork.cold_start(feature_config=FeaturesConfig(...), ...)

**After — flat, no internal classes needed**
BayesianNeuralNetwork.cold_start(
    n_features=2,                    # total columns
    categorical_features={1: 3},    # {col_index: vocab_size}
    ...
)

pybandits/model.py
AnastasiiaKabeshova: applied

docs/src/tutorials/bnn.ipynb
AnastasiiaKabeshova: Cell IDs cannot be stripped — nbformat 5.1+ requires the id field on every cell, so even if nbstripout removes them, nbformat silently re-adds them on the next run, causing the infinite-modification loop.

pybandits/model.py
```diff
@@ -1297,23 +1570,49 @@ def forward_pass(
...
+            Pre-sampled embedding vectors, one array per categorical feature, each of
+            shape ``(n_samples, emb_dim)``. When ``None`` (default) embeddings are
+            sampled fresh on every call.
 
         Returns
         -------
         List[ProbabilityWeight]
             Each element is (probability, weighted_sum) per sample.
         """
+        if self.feature_config.has_categorical:
```
ronshiff1: wrap this in a function please

pybandits/model.py
```diff
@@ -1297,23 +1570,49 @@ def forward_pass(
...
+            _context = np.atleast_2d(context)
+            numerical_arr, cat_indices_dict = self._prepare_context_arrays(_context)
+            parts = [numerical_arr] if numerical_arr.shape[1] > 0 else []
+            for i in range(len(self.feature_config.categorical_features_configs)):
+                if sampled_embedding_vectors is not None:
```
ronshiff1: the weights are always sampled outside.
here you  sample the embedding vectors if they are not given.
I think we should be consistent.
I would sample it outside like we did for the weights - even enhance the sampling function to sample both weights and embedding vectors.
it can be reused later in the QBNN

pybandits/quantitative_model.py
```diff
@@ -1213,43 +1230,64 @@ def cold_start(
...
         """
         n_samples = len(context)
         n_outputs = len(get_args(get_args(get_type_hints(self._to_quantitative_probabilities)["return"])[0]))
 
         result = []
         for sample_idx in range(n_samples):
             sampled_weights_idx = self.bnn._extract_weights(sampled_weights, sample_idx)
+            # Slice the pre-sampled embeddings for this specific sample.
+            # Shape per feature becomes (1, emb_dim) to match forward_pass's n_samples=1 expectation.
+            emb_vectors_idx: Optional[List[np.ndarray]] = (
```
ronshiff1: I suggest to enhance _extract_weights and put this code there

tests/test_quantitative_model.py
```diff
@@ -306,6 +306,19 @@ def test_sample_proba_returns_valid_probabilities_smab(
             assert 0 <= prob <= 1
 
 
+def _make_qbnn_with_categoricals(dimension: int = 1):
```
ronshiff1: why do we need this?
it's just:
 QuantitativeBayesianNeuralNetwork.cold_start(
        dimension=dimension,
        categorical_features={0: 3},

pybandits/quantitative_model.py
```diff
@@ -1174,6 +1174,7 @@ def cold_start(
         cls,
         dimension: PositiveInt = 1,
         n_features: NonNegativeInt = 1,
+        categorical_features: Optional[Dict[int, int]] = None,
```
ronshiff1: Optional[Dict[NonNegativeInt, NonNegativeInt]]

pybandits/quantitative_model.py
```diff
@@ -1185,7 +1186,15 @@ def cold_start(
         dimension : PositiveInt
             Dimension of the quantity (action) space. Default is 1.
         n_features : NonNegativeInt
-            Number of context dimensions (input features in addition to the quantity). Default is 1.
+            Number of numerical context features (used when ``categorical_features`` is ``None``).
```
ronshiff1: isn't this a bug (at least in the documentation...)? 
from BNN cold_start:
       " n_features : PositiveInt
            Total number of columns in the context array, including any categorical columns."
and you send to BNN  n_features = n_features + dimension

pybandits/model.py
```diff
@@ -777,25 +785,151 @@ def _get_distribution_specific_params(cls, shape: Tuple[PositiveInt, ...]) -> Di
...
+    Columns can appear in any order. Categorical features are identified by their
+    explicit ``column_index``; all remaining columns are treated as numerical.
+
+    Parameters
+    ----------
+    n_features : int
+        Total number of columns in the input numpy array. Default 0.
+    categorical_features_configs : List[CategoricalFeatureConfig]
+        List of categorical feature configurations. Each entry carries a
+        ``column_index`` indicating its position in the input array.
```
ronshiff1: Each entry carries CategoricalFeatureConfig, no? or I'm missing something

pybandits/model.py
ronshiff1: I didn't see any change... please don't resolve if you don't do a change.
you're welcome of course to disagree - but please reply - and let's decide what to do

pybandits/model.py
```diff
@@ -1158,83 +1366,149 @@ def model_post_init(self, __context: Any) -> None:
...
+        input_parts : list of PyMC tensors
+            Concatenation of numerical and embedded-categorical inputs.
+        bnn_output : PyMC-compatible array
+            Observed binary targets (``Data`` node or ``Minibatch`` slice).
+        n_total : Optional[int]
+            Full dataset size; passed as ``total_size`` to ``Bernoulli`` when using
+            minibatch (scales the likelihood correctly). ``None`` for full-batch.
+        """
+        next_layer_input = input_parts[0] if len(input_parts) == 1 else math.concatenate(input_parts, axis=1)
+        linear_transform = self._build_bnn_layers(next_layer_input)
```
ronshiff1: it should be something like 
_logit = self._apply_bnn_layers(network_input)

pybandits/cmab.py
```diff
@@ -260,12 +262,26 @@ def update_old_state(
...
+                ).apply_version_adjusted_method("model_dump", "dict")
+                action_state["feature_config"] = feature_config
+
                 action_state.pop("alpha")
                 action_state.pop("betas")
 
+            # 3.0 <= version < 5.0: model_params present but feature_config absent (numerical-only).
+            elif "model_params" in action_state and "feature_config" not in action_state:
+                layer_params = action_state["model_params"]["bnn_layer_params"]
+                n_features = len(layer_params[0][BaseBayesianNeuralNetwork.weight_var_name]["mu"])
```
shaharbar1: Instead of "mu", take the first key from BaseLocationScaleArray.param_map (remove _ from _param_map)

pybandits/model.py
```diff
@@ -301,6 +302,8 @@ class BaseLocationScaleArray(PyBanditsBaseModel, ABC):
     _sigma_array: np.ndarray = PrivateAttr()
     _params: Dict[str, np.ndarray] = PrivateAttr()
     _pymc_class: ClassVar[type]
+    _scipy_sampler: ClassVar[Callable]
```
shaharbar1: Just call it _sampler, no need for the "scipy"

pybandits/model.py
```diff
@@ -352,7 +355,9 @@ def with_dist_parameters(self, **kwargs) -> "BaseLocationScaleArray":
         else:
             raise ValueError(f"Unsupported pydantic version: {pydantic_version}")
 
-    @abstractmethod
+    def _to_scipy_kwargs(self, params: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
```
shaharbar1: _to_sampler_kwargs

pybandits/model.py
```diff
@@ -367,7 +372,28 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+        ----------
+        indices : np.ndarray of shape (n,) with dtype int
+            Row indices to sample from.
+
+        Returns
+        -------
+        np.ndarray of shape (n, ncols)
+        """
+        sliced = {k: v[indices] for k, v in self._params.items()}
+        return self._scipy_sampler(**self._to_scipy_kwargs(sliced))
```
shaharbar1: Great!

pybandits/model.py
shaharbar1: I just wouldn't put a default here.

pybandits/model.py
shaharbar1: I'm not gonna fight over this one, this is my final take:

Image

Your call...

pybandits/model.py
shaharbar1: Please keep it as is.

pybandits/model.py
```diff
@@ -1022,29 +1205,26 @@ def get_layer_params_name(cls, layer_ind: PositiveInt) -> Tuple[str, str]:
...
+            ``EmbeddingParams`` are created when ``feature_config.categorical_features_configs`` is non-empty.
         hidden_dim_list : Optional[List[PositiveInt]]
-            A list of integers specifying the number of hidden units in each hidden layer.
-            If None, no hidden layers are added.
+            Number of hidden units per hidden layer. If None, no hidden layers are added.
         use_layerwise_scaling : bool
             Whether to use layerwise scaling in the network (default is False).
-        dist_class : type[BaseLocationScaleArray], optional
-            The distribution class to use (StudentTArray or NormalArray), by default StudentTArray.
+        dist_type : Literal["normal", "studentt"]
```
shaharbar1: I would suggest to revert this one. create_model_params is less of an outer interface function. It should get finalized classes, rather than literals...

pybandits/model.py
```diff
@@ -1071,46 +1253,72 @@ def create_model_params(
             b_param = dist_class.cold_start(shape=output_dim, **dist_params_init)
             layer_params_init.append(BnnLayerParams(weight=w_param, bias=b_param))
 
-        return BnnParams(bnn_layer_params=layer_params_init)
+        embedding_params = None
+        if feature_config.categorical_features_configs:
+            embedding_params = EmbeddingParams.cold_start(
+                feature_config=feature_config,
+                dist_type=dist_type,
+                **dist_params_init,
```
shaharbar1, line 1262: else:
embedding_params = None

pybandits/model.py
```diff
@@ -1071,46 +1253,72 @@ def create_model_params(
...
-        context : pandas DataFrame of shape (n_samples, n_features)
-            Matrix of contextual features.
+        context : np.ndarray
+            Matrix of contextual features of shape ``(n_samples, n_cols)``.
         """
+        expected_cols = self.feature_config.n_features
+
         try:
-            n_cols_context = np.array(context).shape[1]
+            context_arr = np.array(context)
```
shaharbar1: If the input is an np.array, why do you cast it again to np.array? this is memory-inefficient.

pybandits/model.py
shaharbar1: I agree with Ron. Having the if else with code repeating itself is bad. The issue should be solved by data preparation for the update model, not by different graph building snippets.

.pre-commit-config.yaml
```diff
@@ -27,4 +27,7 @@ repos:
     rev: 0.8.1
     hooks:
       - id: nbstripout
-        args: ['--extra-keys=metadata.kernelspec metadata.language_info.version metadata.language_info.pygments_lexer']
```
shaharbar1, line 30: You can revert this if it is not helping with "id" keys.

shaharbar1 (changes requested):
Nice progress @AnastasiiaKabeshova . Still some way to go to finalize this.

pybandits/model.py
ronshiff1: I think that _finalize_pymc_graph is redundant, 
I would keep the original function's structure, but add a prepare_bnn_input_ouptut function
which returns input_parts, bnn_output
and the function continues from there. 
let's discuss it
The core goal of this function is to create the BNN - with current implementation - it is divided to 3 functions and is less readable.
I would put the input/output creation "hidden" while keeping the BNN implementation cleaerer

AnastasiiaKabeshova: thanks !
it is definitely cleaner that way

pybandits/model.py
AnastasiiaKabeshova: no fights please 
let's put cardinality here :)

pybandits/model.py
AnastasiiaKabeshova: removed

.pre-commit-config.yaml
AnastasiiaKabeshova: i didn't remove it, just refactored a bit

pybandits/model.py
ronshiff1: Please do the following fixes for the docstring:
" Samples probabilities and **logits** from the prior predictive distribution."

remove this:
" This is a refactored version of `sample_proba` that separates weight sampling
        from the forward pass computation."

context: np.ndarray

Returns....
....
List[ProbabilityWeight]
            Each element is a tuple containing the probability of a positive reward and the network logit

pybandits/quantitative_model.py
```diff
@@ -1185,7 +1186,15 @@ def cold_start(
         dimension : PositiveInt
             Dimension of the quantity (action) space. Default is 1.
         n_features : NonNegativeInt
-            Number of context dimensions (input features in addition to the quantity). Default is 1.
+            Total number of columns in the context array, including any categorical columns.
+            Default is 1.
+        categorical_features : Optional[Dict[int, int]]
```
shaharbar1: Fix to NonNegativeInt.

pybandits/quantitative_model.py
```diff
@@ -1213,43 +1230,57 @@ def cold_start(
...
+            Passed as-is to ``forward_pass`` to guarantee a deterministic forward pass per
+            sample — the embeddings are fixed for all quantity evaluations of that sample.
         """
         n_samples = len(context)
         n_outputs = len(get_args(get_args(get_type_hints(self._to_quantitative_probabilities)["return"])[0]))
 
         result = []
         for sample_idx in range(n_samples):
-            sampled_weights_idx = self.bnn._extract_weights(sampled_weights, sample_idx)
+            weights_idx, emb_idx = self.bnn._extract_sample(sampled_weights, sampled_embeddings, sample_idx)
```
shaharbar1: change _extract_sample to extract_sample

pybandits/quantitative_model.py
ronshiff1: Optional[Dict[NonNegativeInt, NonNegativeInt]
Are you sure all this text below is needed?

pybandits/quantitative_model.py
```diff
@@ -1275,13 +1306,27 @@ def sample_proba(self, context: np.ndarray) -> List[QuantitativeProbabilityWeigh
...
             A list of (probability, weight) callables per sample, each taking a quantity (Union[float, np.ndarray]).
         """
-
         _context = np.atleast_2d(context)
         n_samples = _context.shape[0]
 
         sampled_weights = self.bnn.sample_weights(n_samples)
 
-        result = self._to_quantitative_probabilities(context=_context, sampled_weights=sampled_weights)
+        # Pre-sample embedding vectors once so that every call to a returned closure
```
ronshiff1: I would keep only this comment "Build a dummy full array so
        # sample_embeddings can extract the right columns from the context part"

you can add a comment before the 
 sampled_weights = self.bnn.sample_weights(n_samples)
and  sampled_embeddings

"pre-sample weights and embedding vectors so bnn forward is deterministic.
or something like that "

tests/test_model.py
```diff
@@ -957,26 +956,29 @@ def test_can_init_bayesian_neural_network_cc(
 ):
     # at least one beta must be specified
     dim_list = [n_features] + hidden_dim_list
+    fc = FeaturesConfig(n_features=n_features, categorical_features_configs=[])
```
shaharbar1: Can't it be just FeaturesConfig(n_features=n_features)?

pybandits/model.py
```diff
@@ -367,7 +372,28 @@ def sample_rvs(self, size: Tuple[int, ...]) -> np.ndarray:
...
+        ``O(len(indices) × nrows × ncols)``.
+
+        Parameters
+        ----------
+        indices : Union[List[NonNegativeInt], np.ndarray] of shape (n,) with dtype int
+            Row indices to sample from.
+
+        Returns
+        -------
+        np.ndarray of shape (n, ncols)
```
shaharbar1, line 393: Write:
np.ndarray of shape (n, ncols)
     sampled instances

pybandits/model.py
```diff
@@ -644,6 +670,8 @@ class StudentTArray(BaseLocationScaleArray):
 
     _nu_array: np.ndarray = PrivateAttr()
     _pymc_class: ClassVar[type] = PymcStudentT
+    _sampler: ClassVar[Callable] = t.rvs
+    param_map: ClassVar[Dict[str, str]] = {"mu": "loc", "sigma": "scale", "nu": "df"}
```
shaharbar1: = {**BaseLocationScaleArray.param_map,  "nu": "df"}

pybandits/model.py
```diff
@@ -1022,29 +1203,29 @@ def get_layer_params_name(cls, layer_ind: PositiveInt) -> Tuple[str, str]:
     @classmethod
     def create_model_params(
         cls,
-        n_features: PositiveInt,
+        feature_config: FeaturesConfig,
         hidden_dim_list: Optional[List[PositiveInt]],
         use_layerwise_scaling: bool = False,
-        dist_class: type[BaseLocationScaleArray] = StudentTArray,
+        dist_class: type = StudentTArray,
```
shaharbar1: Why did you remove type[BaseLocationScaleArray]?

pybandits/model.py
```diff
@@ -1551,20 +1965,27 @@ def cold_start(
         Self
             An instance of the Bayesian Neural Network initialized with the specified parameters.
         """
+        cat_configs = [
+            CategoricalFeatureConfig(
+                column_index=col_idx,
+                cardinality=cardinality,
+                embedding_dim=ceil(cardinality / 4),
```
shaharbar1: make 4 a class var.

shaharbar1 (changes requested):

shaharbar1 (approved):

ronshiff1 (approved):

PR #126, opened 2026-03-09
Improve OPE function estimator for imbalanced data

shaharbar1:
### Changes:
* Added `calibrate` flag to `_FunctionEstimator` to apply Platt scaling (5-fold `CalibratedClassifierCV`) after fitting for all model types except
  `logreg`.
* Added `class_weight` to Optuna search space for `rf` and `logreg`, and `scale_pos_weight` for `xgb` to handle class imbalance during training.
* Switched Optuna objective scoring from accuracy to `roc_auc` (binary) or `neg_log_loss` (multiclass), per class count.
* Disabled calibration for importance weights estimator (balanced dataset by construction).
* Optimized OPE memory consumption.

## Summary by CodeRabbit

* **New Features**
  * Optional model calibration enabled by default; evaluation and update calls now accept logged data at call time.

* **Improvements**
  * Configurable bootstrap batching (bootstrap_batch_size) to reduce memory during confidence-interval estimation.
  * Data-driven evaluation flow with safer validation, better metric selection for binary vs multiclass, and more memory-efficient multiprocessing.
  * Adjusted sample selection in replay-style estimation affecting which samples contribute.

* **Tests**
  * Updated tests to exercise new validation and evaluation flows.

* **Chores**
  * Version bumped to 4.5.1

### DIFF
--- a/pybandits/offline_policy_estimator.py
+++ b/pybandits/offline_policy_estimator.py
@@ -55,11 +55,18 @@ class BaseOfflinePolicyEstimator(PyBanditsBaseModel, ABC):
         Number of bootstrap samples for confidence interval estimation.
     random_state : int, default=None
         Random seed for bootstrap sampling.
+    bootstrap_batch_size : Optional[int], default=1
+        Number of resamples to process per batch in the bootstrap call.
+        Memory usage is O(bootstrap_batch_size * n_samples). When None,
+        all resamples are processed in a single batch which can cause OOM
+        kills for large datasets. Set to a smaller value (e.g. 1) to
+        reduce peak memory usage at the cost of slightly more overhead.
     """
 
     alpha: Float01 = 0.05
     n_bootstrap_samples: int = 10000
     random_state: Optional[int] = None
+    bootstrap_batch_size: Optional[PositiveInt] = 1
     name: ClassVar
 
     @classmethod
@@ -147,6 +154,7 @@ class BaseOfflinePolicyEstimator(PyBanditsBaseModel, ABC):
             statistic=np.mean,
             confidence_level=1 - self.alpha,
             n_resamples=self.n_bootstrap_samples,
+            batch=self.bootstrap_batch_size,
             random_state=self.random_state,
         )
         low, high = bootstrap_result.confidence_interval
@@ -184,7 +192,12 @@ class ReplayMethod(BaseOfflinePolicyEstimator):
         Number of bootstrap samples for confidence interval estimation.
     random_state : int, default=None
         Random seed for bootstrap sampling.
-
+    bootstrap_batch_size : Optional[int], default=1
+        Number of resamples to process per batch in the bootstrap call.
+        Memory usage is O(bootstrap_batch_size * n_samples). When None,
+        all resamples are processed in a single batch which can cause OOM
+        kills for large datasets. Set to a smaller value (e.g. 1) to
+        reduce peak memory usage at the cost of slightly more overhead.
     """
 
     name = "rep"
@@ -210,10 +223,11 @@ class ReplayMethod(BaseOfflinePolicyEstimator):
             Estimated sample rewards.
         """
         n_samples = action.shape[0]
-        matched_evaluation_policy = estimated_policy[np.arange(n_samples), action]
-        matched_action = matched_evaluation_policy == 1
+        policy_action_prob = estimated_policy[np.arange(n_samples), action]
         sample_reward = (
-            reward * matched_action / matched_action.mean() if matched_action.any() else np.zeros_like(matched_action)
+            reward * policy_action_prob / policy_action_prob.mean()
+            if policy_action_prob.any()
+            else np.zeros_like(reward)
         )
         return sample_reward
 
@@ -235,6 +249,12 @@ class GeneralizedInverseProbabilityWeighting(BaseOfflinePolicyEstimator, ABC):
         Number of bootstrap samples for confidence interval estimation.
     random_state : int, default=None
         Random seed for bootstrap sampling.
+    bootstrap_batch_size : Optional[int], default=1
+        Number of resamples to process per batch in the bootstrap call.
+        Memory usage is O(bootstrap_batch_size * n_samples). When None,
+        all resamples are processed in a single batch which can cause OOM
+        kills for large datasets. Set to a smaller value (e.g. 1) to
+        reduce peak memory usage at the cost of slightly more overhead.
     """
 
     @abstractmethod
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pybandits"
-version = "4.4.0"
+version = "4.5.1"
```
AnastasiiaKabeshova: update the version

pybandits/offline_policy_evaluator.py
```diff
@@ -278,14 +300,21 @@ def _objective(self, trial: Trial, feature_set: np.ndarray, label: np.ndarray) -
...
         -------
         score : float
             The score to be maximized by Optuna.
         """
-        params = self._sample_parameter_space(trial)
+        n_classes = len(np.unique(label))
+        params = self._sample_parameter_space(trial, n_classes=n_classes)
         model = self._model_mapping[self.estimator_type](**params)
-        score = cross_val_score(model, feature_set, label).mean()
+        scoring = "roc_auc" if n_classes == 2 else "neg_log_loss"
```
AnastasiiaKabeshova, line 359: When training data is imbalanced and a rare action appears fewer than 5 times, one CV fold ends up with only one class. roc_auc returns nan for that fold. It happens to me all the time :)
Optuna treats it silently as a valid score
  → arbitrary model selected, no exception raised.
you should handled it in the code

pybandits/offline_policy_evaluator.py
```diff
@@ -345,7 +374,10 @@ def fit(self, X: dict, y: np.ndarray) -> Self:
             model_parameters = self._optimize(feature_set, y, study)
 
         model = self._model_mapping[self.estimator_type](**model_parameters)
-        model.fit(feature_set, y)
+        if self.calibrate and self.estimator_type != "logreg":
+            model = CalibratedClassifierCV(model, cv=5, method="sigmoid").fit(feature_set, y)
```
AnastasiiaKabeshova: Similar here:
cv=5 requires at least 5 samples per class. With a rare action, raises:
  ValueError
It can crash here.

pybandits/offline_policy_estimator.py
```diff
@@ -209,9 +222,7 @@ def estimate_sample_rewards(
         sample_reward : np.ndarray
             Estimated sample rewards.
         """
-        n_samples = action.shape[0]
-        matched_evaluation_policy = estimated_policy[np.arange(n_samples), action]
-        matched_action = matched_evaluation_policy == 1
+        matched_action = action == estimated_policy.argmax(axis=1)
```
AnastasiiaKabeshova: The fix using argmax is correct in principle, but: np.argmax will return the lowest index. When two actions have equal estimated probability (cold-start for ex), the first action by index always wins. 
Should be documented at least?

pybandits/offline_policy_evaluator.py
```diff
@@ -229,18 +247,19 @@ def _sample_parameter_space(self, trial: Trial) -> Dict[str, Union[str, int, flo
         elif self.estimator_type == "rf":
             return {
                 "max_depth": trial.suggest_int("max_depth", 2, 5),
-                "criterion": trial.suggest_categorical("criterion", ["gini", "entropy"]),
+                "criterion": "entropy",
                 "max_features": trial.suggest_int("max_features", 1, 3),
                 "n_estimators": trial.suggest_int("n_estimators", 10, 50),
+                "class_weight": trial.suggest_categorical("class_weight", ["balanced", None]),
                 "n_jobs": -1,
```
AnastasiiaKabeshova, line 258: here you keep "n_jobs": -1, but in next
> elif self.estimator_type == "logreg":

you dropped it => Inconsistency

pybandits/offline_policy_evaluator.py
```diff
@@ -259,13 +278,16 @@ def _sample_parameter_space(self, trial: Trial) -> Dict[str, Union[str, int, flo
                 "reg_lambda": trial.suggest_float("reg_lambda", 0.0, 1.0),
                 "n_jobs": -1,
             }
+            if n_classes == 2:
+                params["scale_pos_weight"] = trial.suggest_float("scale_pos_weight", 1.0, 100.0, log=True)
```
AnastasiiaKabeshova: scale_pos_weight should equal neg_count / pos_count for the training data. Hardcoding [1, 100] assumes the positive class is always the minority and never more than 100:1 imbalanced. 
What about OPE with many actions? the actual ratio could be outside this range.

pybandits/offline_policy_evaluator.py
```diff
@@ -1089,12 +1134,13 @@ def evaluate(
...
             if self._check_argument_required_by_estimators("reward", self.ope_estimators):
-                kwargs["reward"] = self._test_data["reward"][:, self.reward_feature.index(reward_feature)]
+                kwargs["reward"] = test_data["reward"][:, self.reward_feature.index(reward_feature)]
             if self._check_argument_required_by_estimators("expected_reward", self.ope_estimators):
-                kwargs["expected_reward"] = self._estimated_expected_reward[reward_feature]
+                kwargs["expected_reward"] = estimated_expected_reward[reward_feature]
 
             # Summarize policy values and their confidence intervals estimated by OPE estimators
             for ope_estimator in self.ope_estimators:
+                logger.info(ope_estimator)
```
AnastasiiaKabeshova: do you need it ? was it for debug?

pybandits/offline_policy_estimator.py
```diff
@@ -55,11 +55,18 @@ class BaseOfflinePolicyEstimator(PyBanditsBaseModel, ABC):
         Number of bootstrap samples for confidence interval estimation.
     random_state : int, default=None
         Random seed for bootstrap sampling.
+    bootstrap_batch_size : Optional[int], default=1
+        Number of resamples to process per batch in the bootstrap call.
+        Memory usage is O(bootstrap_batch_size * n_samples). When None,
+        all resamples are processed in a single batch which can cause OOM
+        kills for large datasets. Set to a smaller value (e.g. 100) to
```
AnastasiiaKabeshova: in defaults it is 1, not 100

pybandits/offline_policy_evaluator.py
```diff
@@ -124,6 +129,7 @@ class _FunctionEstimator(PyBanditsBaseModel, ClassifierMixin, arbitrary_types_al
     study_name: Optional[str] = None
     multi_action_prediction: bool
     include_action_in_features: bool = True
+    calibrate: bool = True
```
AnastasiiaKabeshova, line 132: This parameter is never testes in tests
(same as roc_auc, class_weight, scale_pos_weight)

was it your intention?

AnastasiiaKabeshova (changes requested):

pybandits/offline_policy_estimator.py
shaharbar1: Fixed this. It was a bad logic! Thanks for pointing this out.

pybandits/offline_policy_evaluator.py
shaharbar1: _FunctionEstimator is not directly tested. All tests are for the OPE class. So it is tested, just not directly...

pybandits/offline_policy_evaluator.py
shaharbar1: I dropped it for logreg beacuse it is not relevant there... For RF, it does matter :-)

pybandits/offline_policy_evaluator.py
shaharbar1: Great catch. checkout the fixed code.

pybandits/offline_policy_evaluator.py
shaharbar1: Fixed. Great catch.

AnastasiiaKabeshova (approved):

PR #129, opened 2026-03-25
BNN Migration from PyMC to NumPyro

shaharbar1:
### Changes:`.
* Replaced PyMC imports with NumPyro in model files, including distribution classes and optimization methods.
* Updated logging to intercept NumPyro logger instead of PyMC.
* Added migration logic for `update_kwargs` to align with NumPyro's requirements.
* Updated tests to reflect changes in model updates and dependencies.

## Summary by CodeRabbit

* **New Features**
  * Early stopping for model training with configurable patience and tolerance.

* **Refactor**
  * Core inference backend migrated to JAX/NumPyro (VI and MCMC); updated training/config semantics (top-level num_steps, optimizer step_size, nuts.target_accept_prob).
  * Model API updated (update-model factory now returns a callable; improved serialization/pickling support).

* **Chores**
  * Package version bumped to 6.0.0; removed PyMC dependency, added NumPyro.

* **Tests**
  * Test suite updated to use new update stubbing utilities (mock_update/apply_mock_update) and adjusted expectations.


pybandits/model.py
```diff
@@ -982,11 +982,50 @@ def validate_inputs(cls, values):
...
+    ----------
+    patience : int
+        Number of consecutive non-improving epochs required before stopping.
+    tolerance : float
+        Threshold for convergence.
+    diff : str
+        Type of difference to check: "relative" or "absolute".
+    """
+
+    _EPSILON: ClassVar[float] = 1e-10
```
ronshiff1: _EPSILON, tolerance,  _previous_loss, should be PositiveFloat 
_no_improvement_count should be NonNegativeInt
I would rename "diff" to "diff_type"
also _EPSILON is missing in docstring

pybandits/model.py
```diff
@@ -982,11 +982,50 @@ def validate_inputs(cls, values):
...
+    diff: Literal["relative", "absolute"] = "relative"
+    _previous_loss: Optional[float] = PrivateAttr(default=None)
+    _no_improvement_count: int = PrivateAttr(default=0)
+
+    def should_stop(self, loss: float) -> bool:
+        """Check if training should stop based on loss convergence."""
+        if self._previous_loss is not None:
+            if self.diff == "relative":
+                change = abs((loss - self._previous_loss) / (abs(self._previous_loss) + self._EPSILON))
+            else:
```
ronshiff1: I would use here "elif self.diff == "absolute":

pybandits/model.py
```diff
@@ -1056,21 +1097,18 @@ class BaseBayesianNeuralNetwork(Model, ABC):
...
+    _supported_optimizers: ClassVar[dict] = {
+        "sgd": noptim.SGD,
+        "momentum": noptim.Momentum,
+        "adagrad": noptim.Adagrad,
+        "adam": noptim.Adam,
+        "clipped_adam": noptim.ClippedAdam,
+    }
+    _jax_activations: ClassVar[dict] = {
+        "tanh": jnp.tanh,
+        "relu": _jax_relu,
```
ronshiff1: why do we sometimes use the function directly, and sometimes wrap it as in 
def _jax_gelu(x):
    """GELU activation function for JAX."""
    return jax.nn.gelu(x)

pybandits/model.py
```diff
@@ -1163,38 +1211,29 @@ def _prepare_context_arrays(self, context: np.ndarray) -> Tuple[np.ndarray, Dict
...
-        return _optimizer
+    def _get_obj_optimizer(self) -> Any:
+        """Build the optimizer from update_kwargs. Always returns an optimizer (uses default if not specified)."""
+        optimizer_type = self.update_kwargs["optimizer_type"]
+        if optimizer_type not in self._supported_optimizers:
+            raise ValueError(
+                f"Invalid optimizer type: {optimizer_type}. "
+                f"Supported optimizers are: {list(self._supported_optimizers.keys())}"
+            )
+        constructor = self._supported_optimizers[optimizer_type]
```
ronshiff1: not the best name ("constructor")

pybandits/model.py
```diff
@@ -1348,192 +1394,152 @@ def _arrange_update_kwargs(self):
...
 
+    def _init_private_attrs(self) -> None:
+        """Initialize private attributes that are derived from public fields.
+
+        These attributes (activation functions, optimizer, etc.) are excluded from pickling
+        and reconstructed on deserialization.
+        """
+        self._numpy_activation_fn = self._numpy_activations[self.activation]
+        self._jax_activation_fn = self._jax_activations[self.activation]
+        self._rng_key = jax.random.PRNGKey(0)
```
AnastasiiaKabeshova: The RNG key is always initialized to PRNGKey(0). This means:
  - Every model instance starts with the same random state
  - After pickling/unpickling, the RNG resets to 0, losing the prior random state progression
  - There is no way for users to set a seed

  Recommendation: Accept an optional random_seed parameter in cold_start() and __init__, and persist it. On deserialization, either restore the key from state or re-derive from the seed. Consider
  using jax.random.PRNGKey(int(time.time_ns())) as default for non-reproducible runs, or require explicit seed.

pybandits/model.py
```diff
@@ -1843,72 +1853,150 @@ def _update(self, context: np.ndarray, rewards: List[BinaryReward]):
...
+                site_mu = {k.removesuffix("_auto_loc"): v for k, v in params.items() if k.endswith("_auto_loc")}
+                site_sigma = {k.removesuffix("_auto_scale"): v for k, v in params.items() if k.endswith("_auto_scale")}
+            elif vi_method == "fullrank_advi":
+                # AutoMultivariateNormal stores a joint loc vector and scale_tril matrix.
+                # Extract exact marginal means and stds directly (no sampling).
+                loc = params["auto_loc"]
+                scale_tril = params["auto_scale_tril"]
+                marginal_std = jnp.sqrt(jnp.sum(scale_tril**2, axis=1))
+                site_mu = guide._unpack_and_constrain(loc, params)
+                site_sigma = guide._unpack_and_constrain(marginal_std, params)
```
AnastasiiaKabeshova: Better to use some public API methods:
- Use guide.sample_posterior(rng_key, params, sample_shape=(n,)) to get per-site samples. 
- Or Predictive(guide, params=params, num_samples=1)?

pybandits/model.py
AnastasiiaKabeshova, line 2095:
```suggestion
        
        Parameters
```

tests/test_model.py
```diff
@@ -790,13 +763,15 @@ def test_invalid_early_stopping_kwargs(
     epochs=st.integers(min_value=1, max_value=100),
     n=st.integers(min_value=1, max_value=100),
 )
-def test_epochs_and_n_mutually_exclusive(n_features: int, update_method: UpdateMethods, epochs: int, n: int) -> None:
-    """Test that specifying both 'epochs' and 'n' raises ValueError."""
-    with pytest.raises(ValueError, match="Cannot specify both 'epochs' and 'n'"):
+def test_epochs_and_num_steps_mutually_exclusive(
+    n_features: int, update_method: UpdateMethods, epochs: int, n: int
```
AnastasiiaKabeshova:
```suggestion
    n_features: int, update_method: UpdateMethods, epochs: int, num_steps: int
```

pybandits/cmab.py
```diff
@@ -103,7 +102,7 @@ def _extract_element_from_probability_weight(
     @validate_call(config=dict(arbitrary_types_allowed=True))
     def predict(
         self,
-        context: ArrayLike,
```
AnastasiiaKabeshova, line 106: why this change? 
It means that passing lists or DataFrames is not allowed anymore?

shaharbar1: Indeed. It was agreed with MLEs. It is causing issues in ci with latest numpy versions.

tests/test_model.py
shaharbar1: Thx!

pybandits/model.py
shaharbar1: I didn't want to use sampling, as it is less accurate, but I'll check for a better option.

pybandits/model.py
shaharbar1: Good point. LMK what you think on the new implementation.

pybandits/model.py
shaharbar1: Changed to cls_

pybandits/model.py
shaharbar1: Good point! Fixed it.

pybandits/model.py
shaharbar1: Thx!
I didn't put epsilon in the docstring because its a classvar...

pybandits/model.py
shaharbar1: I didn't want to use sampling, as it is less accurate, but I'll check for a better option.

shaharbar1: Fixed. Thx!

pybandits/model.py
```diff
@@ -1348,192 +1391,156 @@ def _arrange_update_kwargs(self):
...
 
-        ``x`` must be a numpy array with ``feature_config.n_features`` columns.
-        Numerical columns are passed through as-is. Categorical columns (identified by their
-        ``column_index``) are modelled with Bayesian embedding matrices inside the PyMC graph.
+        n_layers = len(self.model_params.bnn_layer_params)
+        numerical_indices = self.feature_config.numerical_indices
+        cat_configs = self.feature_config.categorical_features_configs
+        has_embeddings = self.model_params.embedding_params is not None and len(cat_configs) > 0
+
+        def model(x, y):
```
ronshiff1: maybe add typehints here?

pybandits/model.py
```diff
@@ -1348,192 +1391,156 @@ def _arrange_update_kwargs(self):
...
+            embedding_matrices = []
+            if has_embeddings:
+                for i, emb_dist in enumerate(self.model_params.embedding_params.embeddings):
+                    emb = numpyro.sample(self.get_embedding_var_name(i), emb_dist.to_numpyro_distribution())
+                    embedding_matrices.append(emb)
+
+            # Data plate with optional minibatching
+            with numpyro.plate("data", N, subsample_size=batch_size) as idx:
+                x_batch = x[idx] if batch_size is not None else x
+                y_batch = y[idx] if (batch_size is not None and y is not None) else y
```
ronshiff1: y can be None?

AnastasiiaKabeshova (approved):

pybandits/model.py
```diff
@@ -1843,72 +1855,156 @@ def _update(self, context: np.ndarray, rewards: List[BinaryReward]):
...
+                if remaining > 0:
+                    epoch_steps_list.append(remaining)
+                if not epoch_steps_list:
+                    epoch_steps_list = [1]
+
+            # Scale the ELBO by 1/N so the loss is a per-sample average.
+            # This decouples learning rate tuning from dataset size: gradients
+            # have consistent magnitude regardless of N, and the KL prior term
+            # isn't dwarfed by the likelihood when N is large.
+            _scaled_model = numpyro_scale(_model, scale=1.0 / n_samples)
```
ronshiff1: I'm pretty sure that it's not needed. please check.
I dug a little bit. this is what I found:
from documentation:

Image
from chatgpt:

Image

Image


This is the entire chat - it also provides a way to read the "scale" parameter.
https://chatgpt.com/share/69c68a89-24a8-8329-8e78-95d87c7a016b
from numpyro.handlers import trace, seed
tr = trace(seed(model, rng_key)).get_trace(x, y, N=N)
print(tr["obs"]["scale"])

enjoy the dig....

pybandits/model.py
```diff
@@ -1843,72 +1855,156 @@ def _update(self, context: np.ndarray, rewards: List[BinaryReward]):
...
+
+            for epoch_steps in epoch_steps_list:
+                svi_state, epoch_losses = _run_epoch(svi_state, epoch_steps)
+
+                epoch_np = np.array(epoch_losses)
+                all_losses.append(epoch_np)
+                pbar.update(1)
+                pbar.set_postfix(loss=f"{float(epoch_np[-1]):.4f}")
+
+                if self._early_stopping_callback is not None:
```
ronshiff1: since we're doing this on the train set, we can consider doing it on the batch level and not the epoch...
(I guess it depends on how much time it takes for an epoch - but then we maybe need to "smooth" the loss....)

pybandits/model.py
```diff
@@ -1843,72 +1855,156 @@ def _update(self, context: np.ndarray, rewards: List[BinaryReward]):
...
+                    if self._early_stopping_callback.should_stop(float(epoch_np[-1])):
+                        break
+
+            pbar.close()
+            self._approx_history = np.concatenate(all_losses) if all_losses else np.array([])
+            params = svi.get_params(svi_state)
+
+            # Extract per-site posterior mean and std from guide params
+            if vi_method == "advi":
```
ronshiff1: can you please put this beautiful code in a function? :D

pybandits/model.py
shaharbar1: I preferred doing it on epoch level. Batch level felt to noisy to me.

pybandits/model.py
shaharbar1: I am familiar with what plate does. That's already in the code. I'm normalizing by dataset size, not by batch size. This is my reason:
image

pybandits/model.py
shaharbar1: Done. I exported the _update code to multiple functions. Indeed a blessed change. Thx!

PR #130, opened 2026-03-29
Backyard clean: remove BLR, flatten Zooming hierarchy

shaharbar1:
### Changes
 * Remove `BayesianLogisticRegression` and `BayesianLogisticRegressionCC` from `model.py`
 * Flatten zooming hierarchy in `quantitative_model.py`:
   - `Zooming` (was `SmabZoomingModel`) absorbs `BaseSmabZoomingModel` — now the single concrete zooming model for sMAB
   - `ZoomingCC` (was `SmabZoomingModelCC`) retains cost-control variant
   - Remove `BaseCmabZoomingModel`, `CmabZoomingModel`, `CmabZoomingModelCC` — zooming is sMAB-only
 * Update imports and type annotations in `actions_manager.py`, `smab.py`, `cmab.py`
 * Remove `BayesianLogisticRegression` tests and update all zooming references across test suite

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved epoch loss reporting and early-stopping to use per-epoch mean loss.

* **Chores**
  * Bumped package version to 6.0.2.
  * Simplified quantitative model internals.

* **Breaking Changes**
  * Removed legacy logistic‑regression model types from the public API.
  * Renamed/consolidated zooming quantitative model names (public imports updated).

* **Documentation**
  * Tutorials/notebooks updated to use the neural‑network model.

* **Tests**
  * Test suite updated to reflect API and naming changes.

### DIFF
--- a/docs/src/tutorials/cmab_simulator.ipynb
+++ b/docs/src/tutorials/cmab_simulator.ipynb
@@ -24,7 +24,7 @@
     "\n",
     "from pybandits.cmab import CmabBernoulli\n",
     "from pybandits.cmab_simulator import CmabSimulator\n",
-    "from pybandits.model import BayesianLogisticRegression, BnnLayerParams, BnnParams, FeaturesConfig, StudentTArray"
+    "from pybandits.model import BayesianNeuralNetwork, BnnLayerParams, BnnParams, FeaturesConfig, StudentTArray"
    ]
   },
   {
@@ -101,14 +101,14 @@
     "# define action model\n",
     "\n",
     "\n",
-    "def create_blr(n_features, bias_mu, bias_sigma, update_method, update_kwargs):\n",
-    "    \"\"\"Create a BayesianLogisticRegression with given parameters.\"\"\"\n",
+    "def create_bnn(n_features, bias_mu, bias_sigma, update_method, update_kwargs):\n",
+    "    \"\"\"Create a BayesianNeuralNetwork with given parameters.\"\"\"\n",
     "    bias = StudentTArray.cold_start(mu=bias_mu, sigma=bias_sigma, shape=1)\n",
     "    weight = StudentTArray.cold_start(shape=(n_features, 1))\n",
     "    layer_params = BnnLayerParams(weight=weight, bias=bias)\n",
     "    model_params = BnnParams(bnn_layer_params=[layer_params])\n",
     "    feature_config = FeaturesConfig(n_features=n_features)\n",
-    "    return BayesianLogisticRegression(\n",
+    "    return BayesianNeuralNetwork(\n",
     "        model_params=model_params,\n",
     "        feature_config=feature_config,\n",
     "        update_method=update_method,\n",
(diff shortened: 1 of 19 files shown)


pybandits/model.py
```diff
@@ -1909,30 +1909,37 @@ def _svi_body(state, _):
...
                         f"{self._early_stopping_callback.patience} consecutive epochs. "
-                        f"Last loss: {float(epoch_np[-1]):.6f}."
+                        f"Best loss: {best_loss:.6f}, last loss: {epoch_loss:.6f}."
                     )
                     break
 
         pbar.close()
         self._approx_history = np.concatenate(all_losses) if all_losses else np.array([])
-        params = svi.get_params(svi_state)
+        params = svi.get_params(best_svi_state)
```
AnastasiiaKabeshova: restore is active for all BNN training, not just early stopping: Every update() call now silently restores weights from the lowest-loss epoch rather than the final epoch. 
Is this a  behavior we expect for all BNN trainings now?

pybandits/model.py
```diff
@@ -1005,12 +1005,12 @@ def should_stop(self, loss: float) -> bool:
         """Check if training should stop based on loss convergence."""
         if self._previous_loss is not None:
             if self.diff_type == "relative":
-                change = abs((loss - self._previous_loss) / (abs(self._previous_loss) + self._epsilon))
+                change = (self._previous_loss - loss) / (abs(self._previous_loss) + self._epsilon)
```
AnastasiiaKabeshova, line 1001: sign semantics completely reversed here.
Now we count a step toward stopping whenever loss doesn't decrease. It is not what usually done for early stopping I think.
Why don't you keep as it was?

pybandits/model.py
```diff
@@ -2412,24 +2419,3 @@ class BayesianNeuralNetworkMOCC(BaseBayesianNeuralNetworkMO, ModelMO, ModelCC):
     cost : NonNegativeFloat
         Cost associated to the Bayesian Neural Network model.
     """
-
-
-class BayesianLogisticRegression(BayesianNeuralNetwork):
```
AnastasiiaKabeshova, line 2417: do not forget update tutorial notebook after this removal

docs/src/tutorials/cmab_simulator.ipynb
AnastasiiaKabeshova: please undate docstrings

pybandits/model.py
AnastasiiaKabeshova, line 2417: since you hardly removed all those classes, should you bumpe the version to 6.1?

AnastasiiaKabeshova (changes requested):

pybandits/model.py
shaharbar1: Yes. I mean, that's what I thought... Do you see anything wrong about it? Do you think I should have a flag for it?

pybandits/model.py
shaharbar1: I don't think so. I don't see it as a big change.

shaharbar1: Done.

pybandits/model.py
shaharbar1: I added a flag with a default to restore best state.

AnastasiiaKabeshova (approved):

PR #134, opened 2026-04-09
Add centralized seeded RNG

shaharbar1:
### Changes:

* Add random_seed field and private _rng (numpy.random.Generator) to BaseMab — single RNG seeded at construction and propagated to all sample_proba calls
* Propagate rng parameter through sample_proba across BaseModel, BaseModelSO, BaseModelMO, Beta, BayesianNeuralNetwork, SmabZoomingModel, and CmabZoomingModel
  * Beta.sample_proba vectorized via rng.beta() (replaces per-sample betavariate loop)
  * BayesianDist.sample_rvs / sample_at_indices pass rng as random_state to scipy samplers
* Replace np.random.random() with self._rng.random() for quantitative action refinement
* Add __eq__ to PyBanditsBaseModel comparing only serializable fields (excludes _rng)
* Expose random_seed in cold_start() factory method

## Summary by CodeRabbit

* **New Features**
  - Optional random_seed for reproducible/deterministic runs.

* **Improvements**
  - Centralized RNG: sampling now accepts/uses an explicit NumPy generator.
  - Notebooks updated to pass explicit RNG instances for reproducible examples.

* **Breaking Changes**
  - Sampling APIs now require an explicit RNG parameter.
  - Model equality tightened: requires identical runtime types and compares serialized state.

* **Chores**
  - Package version bumped to 6.0.6; tests include a deterministic RNG fixture.


pybandits/model.py
```diff
@@ -532,28 +556,6 @@ def params(self) -> Dict[str, np.ndarray]:
         """
         return self._params
 
-    def __eq__(self, other: Any) -> bool:
```
AnastasiiaKabeshova, line 534: Before __eq__ compared _mu_array and _sigma_array.
Two arrays with different internal state but equal public fields will now compare equal.
Could you maybe add some test to confirm that PyBanditsBaseModel.__eq__ will work correctly as well?

pybandits/model.py
```diff
@@ -1556,14 +1544,17 @@ def model(x: jax.Array, y: jax.Array):
...
+    def sample_weights(self, n_samples: PositiveInt, rng: np.random.Generator) -> List[Tuple[np.ndarray, np.ndarray]]:
         """
         Sample weights and biases for each sample and each layer.
 
         Parameters
         ----------
         n_samples : PositiveInt
             The number of samples (users) to draw weights for. Must be positive.
+        rng : np.random.Generator
+            Numpy random generator forwarded to scipy samplers. Enables reproducible
```
AnastasiiaKabeshova: It is not scipy anymore, it calls numpy now
Please update docstring

pybandits/model.py
```diff
@@ -1596,6 +1587,8 @@ def sample_embeddings(self, context: np.ndarray) -> Optional[List[np.ndarray]]:
         context : np.ndarray
             Context matrix, shape ``(n_samples, feature_config.n_features)``.
             Categorical columns contain integer indices into the embedding vocabulary.
+        rng : np.random.Generator
+            Numpy random generator forwarded to scipy samplers.
```
AnastasiiaKabeshova: same here, update docstring

tests/test_cmab.py
```diff
@@ -1171,3 +1171,31 @@ def test_extract_element_from_probability_weight_unsupported_type(unsupported_ty
...
         BaseCmabBernoulli._extract_element_from_probability_weight(0, unsupported_type)
+
+
+@settings(deadline=None, max_examples=10)
+@given(
+    action_ids=st.lists(st.text(min_size=1), min_size=1, max_size=4, unique=True),
+    n_features=st.integers(min_value=1, max_value=4),
+    random_seed=st.integers(min_value=0, max_value=2**31 - 1),
+)
+def test_random_seed_propagates_to_bnn(action_ids: List[str], n_features: int, random_seed: int) -> None:
```
AnastasiiaKabeshova: this test verifies the seed attribute propagates, but not that running the pipeline twice with the same seed yields identical predictions/actions.

could you add an end-to-end reproducibility test (same seed → same predict output)?

AnastasiiaKabeshova (changes requested):

pybandits/model.py
shaharbar1: But internal states are derived from the public states. I updated existing tests to verify private attrs match public ones. I think it better addresses your comment.

tests/test_cmab.py
shaharbar1: Done.

AnastasiiaKabeshova (approved):

PR #138, opened 2026-05-04
Fix BNN SVI warm-start and minibatching stability

shaharbar1:
### Changes:
* Add `_build_svi_guide_init()` to seed the SVI guide from current model weights — uses `init_to_value` when weights are non-zero, falling back to `init_to_median` on cold start, so updates resume from the existing posterior instead of reinitializing
* Set guide `init_scale` to the mean of current sigma values so the guide's initial spread matches the model's posterior width rather than AutoNormal's fixed 0.1 default
* Skip `numpyro.plate` when `batch_size >= N` or `None` via `nullcontext` to avoid numpyro subsample indexing when the full dataset fits in one batch
* Stop applying `softplus` to `_auto_scale` params on readback — they are already constrained; clamp to `_min_vi_sigma` (1e-6) instead to prevent zero sigmas

## Summary by CodeRabbit

* **Improvements**
  * Enhanced Bayesian Neural Network model training efficiency with optimized batch processing
  * Improved variational inference guide initialization for better model convergence
  * Refined parameter extraction to ensure stable variance estimates during model inference

### DIFF
--- a/pybandits/model.py
+++ b/pybandits/model.py
@@ -22,6 +22,7 @@
 
 import warnings
 from abc import ABC, abstractmethod
+from contextlib import nullcontext
 from copy import deepcopy
 from math import ceil
 from types import ModuleType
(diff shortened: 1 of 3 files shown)


pybandits/model.py
AnastasiiaKabeshova, line 2016: Right now there's no test that directly verifies sigma comes out of _extract_advi_params correctly. Please add a small, direct test:

  - Build a fake params dict with a known _auto_scale value, e.g. 0.01.
  - Call _extract_advi_params.
  - Assert the returned sigma is 0.01 — not softplus(0.01)

AnastasiiaKabeshova (approved):
LGTM.
one non blocking comment about the test

PR #140, opened 2026-05-12
Introduce BaseMetaModel abstraction for cmab actions managers

AnastasiiaKabeshova:
### Summary

Introduces a `BaseMetaModel` layer between `ActionsManager` (sMAB and cMAB) and per-action models. Managers store a `meta_model` field and delegate `update`, `sample_proba`, and `reset` to it. Per-action models live in the meta-model; `actions` is a property over `meta_model.actions_view()` so `mab.actions` and downstream code keep the same model instances.

Default implementation: `PerActionMetaModel` (historical per-action dispatch — **behaviour unchanged**). `SmabMetaModel[T]` and `CmabMetaModel[T]` hold the typed validators formerly on the manager classes. Optional `meta_model=` supports future shared-backbone meta-models.

### Changes

- **New module** `pybandits/meta_model.py`: `BaseMetaModel`, `PerActionMetaModel`, `SmabMetaModel[T]`, `CmabMetaModel[T]`, module aliases for pickling; `SmabModelType` / `CmabModelType` moved here.
- **`pybandits/actions_manager.py`**: `meta_model` as a Pydantic field (built via `_build_meta_model` from `actions=` or passed explicitly); `actions` as a view; base `sample_proba`; sMAB/cMAB `_update_actions` → `meta_model.update`.
- **`pybandits/mab.py`**: `_get_action_probabilities` uses `actions_manager.sample_proba(...)` instead of a local loop.
- **`pybandits/cmab.py`**, **`pybandits/transfer.py`**: serialized state under `actions_manager.meta_model.actions`; cMAB `update_old_state` migrates legacy flat `actions` where applicable.
- **Tests**: new `tests/test_meta_model.py`; integration tests in `tests/test_actions_manager.py`; updates in `tests/test_cmab.py`, `tests/test_transfer.py`.

## Summary by CodeRabbit

* **New Features**
  * Added a meta-model layer for action management.
  * Sampling now supports restricting to valid actions.

* **Bug Fixes**
  * State migration updated to handle legacy and current action layouts.

* **Refactor**
  * Training updates and probability sampling now delegate via the meta-model.
  * Action storage/exposure moved to the meta-model view for consistent wiring.

* **Tests**
  * New/regressed tests verifying meta-model behavior, delegation, and migration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/PlaytikaOSS/pybandits/pull/140?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pybandits/actions_manager.py
+++ b/pybandits/actions_manager.py
@@ -22,8 +22,21 @@
 
 import warnings
 from abc import ABC, abstractmethod
-from collections import defaultdict
-from typing import Any, Callable, ClassVar, Dict, Generic, List, Optional, Set, Tuple, Type, TypeVar, Union
+from typing import (
+    Any,
+    ClassVar,
+    Dict,
+    Generic,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    get_args,
+    get_origin,
+)
 
 import numpy as np
 from numpy.typing import ArrayLike
@@ -32,19 +45,24 @@ from pydantic import (
     Field,
     NonNegativeInt,
     NonPositiveInt,
-    field_validator,
+    PositiveInt,
+    model_validator,
     validate_call,
 )
 
 from pybandits.base import (
-    ACTION_IDS_PREFIX,
-    QUANTITATIVE_ACTION_IDS_PREFIX,
     ActionId,
     BinaryReward,
     PositiveProbability,
     PyBanditsBaseModel,
 )
 from pybandits.base_model import BaseModel, BaseModelMO, BaseModelSO
+from pybandits.meta_model import (
+    BaseMetaModel,
+    CmabPerActionMetaModel,
+    PerActionMetaModel,
+    SampleProbaResult,
+)
 from pybandits.model import (
     BaseBayesianNeuralNetwork,
     BaseBayesianNeuralNetworkMO,
@@ -70,7 +88,16 @@ from pybandits.quantitative_model import (
     Zooming,
     ZoomingCC,
 )
-from pybandits.utils import classproperty, extract_argument_names_from_function
+
+SmabModelType = TypeVar("SmabModelType", bound=Union[BaseBeta, BaseBetaMO, BaseZooming])
+CmabModelType = TypeVar(
+    "CmabModelType",
+    bound=Union[
+        BaseBayesianNeuralNetwork,
+        BaseBayesianNeuralNetworkMO,
+        BaseQuantitativeBayesianNeuralNetwork,
+    ],
+)
 
 
 class ActionsManager(PyBanditsBaseModel, ABC):
@@ -86,35 +113,39 @@ class ActionsManager(PyBanditsBaseModel, ABC):
 
     Parameters
     ----------
-    actions : Dict[ActionId, Model]
-        The list of possible actions, and their associated Model.
+    meta_model : BaseMetaModel
+        The meta-model that owns per-action state and dispatches ``sample_proba``,
+        ``update``, and ``reset``. Constructed automatically from an ``actions``
+        dict when the manager is instantiated via the ``actions=`` kwarg.
     delta : Optional[PositiveProbability]
         The confidence level for the adaptive window. None for skipping the change point detection.
     """
 
-    actions: Dict[ActionId, BaseModel]
+    meta_model: BaseMetaModel
     delta: Optional[PositiveProbability] = None
-    _no_change_point: ClassVar[NonNegativeInt] = -1
-    _min_adaptive_window_size: ClassVar[NonPositiveInt] = 10000
+    _no_change_point: ClassVar[NonPositiveInt] = -1
+    _min_adaptive_window_size: ClassVar[PositiveInt] = 10000
     _memory_parameters_suffix: ClassVar[str] = "_memory"
     actions_with_change: Set[Tuple[ActionId, NonNegativeInt]] = Field(default_factory=set)
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
-    @field_validator("actions", mode="after")
+    @property
+    def actions(self) -> Dict[ActionId, BaseModel]:
+        """Per-action models, delegated to ``meta_model.actions``."""
+        return self.meta_model.actions
+
     @classmethod
-    def at_least_one_action_is_defined(cls, v):
-        # validate number of actions
-        if len(v) == 0:
-            raise AttributeError("At least one action should be defined.")
-        elif len(v) == 1:
-            warnings.warn("Only a single action was supplied. This MAB will be deterministic.")
-        # validate that all actions are of the same configuration
-        action_models = list(v.values())
-        action_type = cls._get_field_type("actions")
-        if any(not isinstance(action, action_type) for action in action_models):
-            raise TypeError(f"All actions should follow {action_type} type.")
-        return v
+    def _get_meta_model_cls(cls) -> Type[BaseMetaModel]:
+        """Return the concrete meta-model class from the manager's ``meta_model`` field annotation."""
+        meta_model_field = cls.model_fields.get("meta_model")
+        if meta_model_field is not None and meta_model_field.annotation is not None:
+            annotation = meta_model_field.annotation
+            type_args = get_args(annotation)
+            if type_args and not isinstance(type_args[0], TypeVar):
+                return annotation  # type: ignore[return-value]
+            return get_origin(annotation) or annotation  # type: ignore[return-value]
+        return PerActionMetaModel
 
     @classmethod
     def _get_expected_memory_length(cls, actions: Dict[ActionId, BaseModel]) -> NonNegativeInt:
@@ -152,13 +183,23 @@ class ActionsManager(PyBanditsBaseModel, ABC):
         quantitative_action_ids: Optional[Set[ActionId]] = None,
         kwargs: Optional[Dict[str, Any]] = None,
         actions_with_change: Optional[Set[Tuple[ActionId, NonNegativeInt]]] = None,
+        meta_model: Optional[BaseMetaModel] = None,
     ):
-        kwargs = kwargs or {}
-        actions = self._instantiate_actions(
-            actions=actions, action_ids=action_ids, quantitative_action_ids=quantitative_action_ids, kwargs=kwargs
-        )
+        action_args = (actions, action_ids, quantitative_action_ids)
+        if meta_model is not None and any(a is not None for a in action_args):
+            raise ValueError(
+                "Provide either 'meta_model' or action-construction arguments "
+                "('actions', 'action_ids', 'quantitative_action_ids'), not both."
+            )
+        if meta_model is None:
+            meta_model = self._get_meta_model_cls()(
+                actions=actions,
+                action_ids=action_ids,
+                quantitative_action_ids=quantitative_action_ids,
+                kwargs=kwargs or {},
+            )
         actions_with_change = actions_with_change or set()
-        super().__init__(actions=actions, delta=delta, actions_with_change=actions_with_change)
+        super().__init__(meta_model=meta_model, delta=delta, actions_with_change=actions_with_change)
 
     def _validate_update_params(
         self,
@@ -202,6 +243,31 @@ class ActionsManager(PyBanditsBaseModel, ABC):
     @classmethod
     def _to_key(cls, key: str) -> str:
         return key.replace(cls._memory_parameters_suffix, "")
+
+    def sample_proba(
+        self,
+        rng: np.random.Generator,
+        valid_action_ids: Optional[Set[ActionId]] = None,
+        **kwargs: Any,
+    ) -> Dict[ActionId, SampleProbaResult]:
+        """Sample per-action probabilities/scores for the bandit's predict path.
+
+        Delegates to ``self.meta_model.sample_proba``. With the default
+        ``PerActionMetaModel`` this is a per-action dispatch loop; alternative
+        meta-models (e.g. a shared backbone) may evaluate all actions jointly.
+
+        Parameters
+        ----------
+        rng : numpy.random.Generator
+            Central random generator from the bandit.
+        valid_action_ids : Optional[Set[ActionId]]
+            If provided, restrict sampling to these action ids; otherwise
+            sample for all actions.
+        **kwargs
+            Forwarded to per-action ``sample_proba`` (e.g. ``context`` for
+            cmab, ``n_samples`` for smab).
+        """
+        return self.meta_model.sample_proba(rng=rng, valid_action_ids=valid_action_ids, **kwargs)
 
     @validate_call(config=dict(arbitrary_types_allowed=True))
     def update(
@@ -276,7 +342,11 @@ class ActionsManager(PyBanditsBaseModel, ABC):
                 for action_model in self.actions.values():
                     if not isinstance(action_model, QuantitativeModel):
                         action_model.reset()
-                regular_actions = [a for a in actions if not isinstance(self.actions[a], QuantitativeModel)]
+                # Non-quantitative models were just reset; retrain on the trimmed memory window.
+                # Filter from `actions_memory` so the action labels align with `rewards_memory`
+                # / `memory_kwargs` (which are filtered from the same source below).
+                regular_actions = [a for a in actions_memory if not isinstance(self.actions[a], QuantitativeModel)]
+                # Quantitative models keep their state (no reset), so they only need the new batch.
                 quantitative_actions = [a for a in actions if isinstance(self.actions[a], QuantitativeModel)]
 
                 if regular_actions:
@@ -647,175 +717,6 @@ class ActionsManager(PyBanditsBaseModel, ABC):
             raise TypeError(f"Model type {type(action_model)} not supported.")
         return current_sum, current_trials
 
-    @classmethod
-    def _instantiate_actions(
-        cls,
-        actions: Optional[Dict[ActionId, Model]],
-        action_ids: Optional[Set[ActionId]],
-        quantitative_action_ids: Optional[Set[ActionId]],
-        kwargs: Dict[str, Any],
-    ):
-        """
-        Utility function to instantiate the action models based on the provided kwargs.
-
-        Parameters
-        ----------
-        actions : Optional[Dict[ActionId, Model]]
-            The list of possible actions and their associated models.
-        action_ids : Optional[Set[ActionId]]
-            The list of possible actions.
-        quantitative_action_ids : Optional[Set[ActionId]]
-            The list of quantitative actions.
-        kwargs : Dict[str, Any]
-            Additional parameters for the mab and for the action model.
-
-        Returns
-        -------
-        actions : Dict[ActionId, Model]
-            Dictionary of actions and the parameters of their associated model.
-        """
-        if actions is None:
-            action_specific_kwargs, quantitative_action_specific_kwargs = cls._extract_action_specific_kwargs(kwargs)
-
-            # Extract inner_action_ids
-            inner_action_ids = action_ids or set(action_specific_kwargs)
-            inner_quantitative_action_ids = quantitative_action_ids or set(quantitative_action_specific_kwargs)
-            if not inner_action_ids and not inner_quantitative_action_ids:
-                raise ValueError("At least one action should be defined.")
-
-            # Assign model for each action
-            (
-                model_cold_start,
-                quantitative_model_cold_start,
-                action_general_kwargs,
-                quantitative_action_general_kwargs,
-            ) = cls._extract_action_model_class_and_attributes(kwargs)
-
-            # Instantiate the actions
-            actions = {}
-            for action_ids, cold_start, general_kwargs, specific_kwargs in zip(
-                [inner_action_ids, inner_quantitative_action_ids],
-                [model_cold_start, quantitative_model_cold_start],
-                [action_general_kwargs, quantitative_action_general_kwargs],
-                [action_specific_kwargs, quantitative_action_specific_kwargs],
-            ):
-                for a in action_ids:
-                    actions[a] = cold_start(**general_kwargs, **specific_kwargs.get(a, {}))
-
-        return actions
-
-    @staticmethod
-    def _extract_action_specific_kwargs(kwargs: Dict[str, Any]) -> Tuple[Dict[str, Dict], Dict[str, Dict]]:
-        """
-        Utility function to extract kwargs that are specific for each action when constructing the action model.
-
-        Parameters
-        ----------
-        kwargs : Dict[str, Any]
-            Additional parameters for the mab and for the action model.
-
-        Returns
-        -------
-        action_specific_kwargs : Dict[str, Dict]
-            Dictionary of actions and the parameters of their associated model.
-        quantitative_action_specific_kwargs : Dict[str, Dict]
-            Dictionary of quantitative actions and the parameters of their associated model.
-        kwargs : Dict[str, Any]
-            Dictionary of parameters and their quantities, without the action_specific_kwargs.
-        """
-        action_specific_kwargs = defaultdict(dict)
-        quantitative_action_specific_kwargs = defaultdict(dict)
-        for keyword in list(kwargs):
-            argument = kwargs[keyword]
-            for prefix, target_kwargs in zip(
-                [ACTION_IDS_PREFIX, QUANTITATIVE_ACTION_IDS_PREFIX],
-                [action_specific_kwargs, quantitative_action_specific_kwargs],
-            ):
-                if keyword.startswith(prefix) and type(argument) is dict:
-                    kwargs.pop(keyword)
-                    inner_keyword = keyword.split(prefix)[1]
-                    for action_id, value in argument.items():
-                        target_kwargs[action_id][inner_keyword] = value
-        return dict(action_specific_kwargs), dict(quantitative_action_specific_kwargs)
-
-    @classmethod
-    def _extract_action_model_class_and_attributes(
-        cls, kwargs
-    ) -> Tuple[Callable, Callable, Dict[str, Dict], Dict[str, Dict]]:
-        """
-        Utility function to extract kwargs that are specific for each action when constructing the action model.
-
-        Parameters
-        ----------
-        kwargs : Dict[str, Any]
-            Additional parameters for the mab and for the action model.
-
-        Returns
-        -------
-        action_model_cold_start : Callable
-            Function handle for factoring the required action model.
-        quantitative_action_model_cold_start : Callable
-            Function handle for factoring the required quantitative action model.
-        action_general_kwargs : Dict[str, any]
-            Dictionary of parameters and their values for the action model.
-        quantitative_action_general_kwargs : Dict[str, any]
-            Dictionary of parameters and their values for the quantitative action model.
-        """
-        action_model_classes = cls._action_model_classes
-        if len(action_model_classes) > 2:
-            raise ValueError("Only up to two types of action models are supported.")
-        quantitative_model_cold_start = model_cold_start = lambda **kwargs: None  # dummy callable
-        action_general_kwargs = quantitative_action_general_kwargs = None
-        for action_model_class in action_model_classes:
-            if hasattr(action_model_class, "cold_start"):
-                action_model_cold_start = action_model_class.cold_start
-                action_model_attributes = extract_argument_names_from_function(action_model_cold_start)
-                # cover for cold_start kwargs
-                action_model_attributes = action_model_attributes + extract_argument_names_from_function(
-                    action_model_class
-                )
-            else:
-                action_model_cold_start = action_model_class
-                action_model_attributes = extract_argument_names_from_function(action_model_cold_start)
-            general_kwargs = {k: kwargs[k] for k in action_model_attributes if k in kwargs.keys()}
-
-            if issubclass(action_model_class, (Model, ModelMO)):
-                model_cold_start = action_model_cold_start
-                action_general_kwargs = general_kwargs
-            elif issubclass(action_model_class, QuantitativeModel):
-                quantitative_model_cold_start = action_model_cold_start
-                quantitative_action_general_kwargs = general_kwargs
-            else:
-                raise TypeError(f"Unsupported action model class: {action_model_class}")
-
-        used_kwargs = (action_general_kwargs or {}) | (quantitative_action_general_kwargs or {})
-        for k in used_kwargs.keys():
-            kwargs.pop(k)
-
-        return (
-            model_cold_start,
-            quantitative_model_cold_start,
-            action_general_kwargs,
-            quantitative_action_general_kwargs,
-        )
-
-    @classproperty
-    def _action_model_classes(cls) -> Tuple[Type[BaseModel], ...]:
-        """
-        Utility function to extract the action model classes from the actions field.
-
-        Returns
-        -------
-        action_model_classes : Tuple[Type[BaseModel],...]
-            Tuple of action model classes.
-        """
-        action_model_type = cls._get_field_type("actions")
-        action_model_classes = action_model_type if isinstance(action_model_type, tuple) else (action_model_type,)
-        return action_model_classes
-
-
-SmabModelType = TypeVar("SmabModelType", bound=Union[BaseBeta, BaseBetaMO, BaseZooming])
-
 
 class SmabActionsManager(ActionsManager, Generic[SmabModelType]):
     """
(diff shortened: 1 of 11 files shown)


pybandits/actions_manager.py
```diff
@@ -950,6 +986,34 @@ class CmabActionsManager(ActionsManager, GenericModel, Generic[CmabModelType]):
     """
 
     actions: Dict[ActionId, CmabModelType]
```
shaharbar1: I thought you would remove 'actions' from actions manager and replace it with the model.

pybandits/actions_manager.py
```diff
@@ -950,6 +986,34 @@ class CmabActionsManager(ActionsManager, GenericModel, Generic[CmabModelType]):
...
+        if self._meta_model is None:
+            self._meta_model = PerActionMetaModel(actions=self.actions)
+
+    @property
+    def meta_model(self) -> BaseMetaModel:
+        """The meta-model owning per-action dispatch for this manager.
+
+        Set automatically from ``actions`` at construction time.
+        """
+        assert self._meta_model is not None, "meta_model not initialised; model_post_init was not called"
```
shaharbar1: assert is for tests, not for code.

pybandits/actions_manager.py
shaharbar1: I mean that's the whole idea, no? to have the meta model as the abstraction of all actions.

shaharbar1: And it would be also for smab, not just for cmab...

pybandits/actions_manager.py
```diff
@@ -215,6 +217,35 @@ def _to_memory_key(cls, key: str) -> str:
     def _to_key(cls, key: str) -> str:
         return key.replace(cls._memory_parameters_suffix, "")
 
+    def sample_proba(
```
shaharbar1, line 247: This is missing in the father class.
Anyhow, this should be reviewed after getting rid of actions and fully moving to meta model.

pybandits/meta_model.py
```diff
@@ -0,0 +1,264 @@
...
+    on both Pydantic v1 and v2 — Pydantic v1's standard field validation
+    deep-copies nested Pydantic models, which would break instance sharing
+    with the enclosing manager.
+    """
+
+    # PrivateAttr storage bypasses Pydantic validation; the dict reference and
+    # the per-action model instances pass through unchanged.
+    _actions: Dict[ActionId, BaseModel] = PrivateAttr(default_factory=dict)
+
+    def __init__(self, **data: Any) -> None:
```
shaharbar1: I don't understand this.
I think 'actions' should be an attribute.

pybandits/meta_model.py
```diff
@@ -0,0 +1,264 @@
...
+            Additional update inputs. For contextual bandits this includes
+            ``context: np.ndarray``.
+        """
+
+    @abstractmethod
+    def reset(self) -> None:
+        """Reset every per-action model to its cold-start state."""
+
+
+class PerActionMetaModel(BaseMetaModel):
```
shaharbar1: I think the abstraction should be general and the two children should be SmabMetaModel and CmabMetaModel.
I don't get why to do abstraction with just one child.

shaharbar1 (changes requested):

pybandits/actions_manager.py
AnastasiiaKabeshova: Good catch, replaced with raise RuntimeError.

pybandits/actions_manager.py
AnastasiiaKabeshova: Base now delegates to meta_model.sample_proba(...)

pybandits/actions_manager.py
AnastasiiaKabeshova: _meta_model + meta_model on base; all dispatch (predict + update) routes through it for smab and cmab

pybandits/meta_model.py
AnastasiiaKabeshova: they'd be identical today; the split becomes meaningful when BackboneHeadsMetaModel is the cmab-specific alternative

pybandits/meta_model.py
AnastasiiaKabeshova: added a comment

pybandits/actions_manager.py
```diff
@@ -98,9 +100,27 @@ class ActionsManager(PyBanditsBaseModel, ABC):
     _min_adaptive_window_size: ClassVar[NonPositiveInt] = 10000
     _memory_parameters_suffix: ClassVar[str] = "_memory"
     actions_with_change: Set[Tuple[ActionId, NonNegativeInt]] = Field(default_factory=set)
+    _meta_model: Optional[BaseMetaModel] = PrivateAttr(default=None)
```
shaharbar1: I think it should be an attribute, not a private. We want it to be serialized and deserialized with the model

pybandits/actions_manager.py
shaharbar1: That's great, but why is actions still here. MetaModel should be the owner of actions. It shouldn't be a member of actions manager anymore (I mean not directly, but only via meta model)

pybandits/meta_model.py
shaharbar1: No, they won't be identical, at least not in type hints. Smab actions would be BaseBeta, etc. and Cmab actions would be BaseBayesianNeuralNetwork, etc.

pybandits/meta_model.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # `actions` is stored as a PrivateAttr (not a Pydantic field) so that the
+    # caller's exact dict reference — and each per-action model instance — is
+    # preserved.  A regular dict field containing nested Pydantic models is
+    # deep-copied during validation, which would silently break the
+    # instance-identity contract the enclosing manager relies on (mutations via
+    # meta_model.update must be visible through manager.actions[action_id], and
+    # change-point detection reads model state from the same objects).
+    # PrivateAttr bypasses that validation entirely, so no copy is ever made.
+    _actions: Dict[ActionId, BaseModel] = PrivateAttr(default_factory=dict)
```
shaharbar1: It should be attribute, not private.

shaharbar1 (changes requested):

pybandits/actions_manager.py
```diff
@@ -802,43 +861,44 @@ def _extract_action_model_class_and_attributes(
     @classproperty
     def _action_model_classes(cls) -> Tuple[Type[BaseModel], ...]:
```
shaharbar1: This, along with _instantiate_actions should move to the MetaModel, as he is incharge of actions now.

pybandits/actions_manager.py
```diff
@@ -931,53 +962,14 @@ class CmabActionsManager(ActionsManager, Generic[CmabModelType]):
 
     Parameters
     ----------
-    actions : Dict[ActionId, BaseBayesianNeuralNetwork]
-        The list of possible actions, and their associated Model.
+    actions : Dict[ActionId, CmabModelType]
```
shaharbar1: Replace action with metamodel. actions is just a property now.

pybandits/cmab.py
```diff
@@ -216,7 +216,16 @@ def update_old_state(cls, state: Dict[str, Serializable]) -> Dict[str, Serializa
         state = super().update_old_state(state)
 
         # Migrate update_kwargs from old PyMC format to new NumPyro format
-        for action_id, action_state in state["actions_manager"]["actions"].items():
+        # Support both old format (actions_manager.actions) and current format (actions_manager.meta_model.actions).
+        actions_manager = cast(Dict, state["actions_manager"])
```
shaharbar1, line 220: why do we need this cast(...) here?

pybandits/meta_model.py
```diff
@@ -0,0 +1,391 @@
...
+            Additional update inputs. For contextual bandits this includes
+            ``context: np.ndarray``.
+        """
+
+    @abstractmethod
+    def reset(self) -> None:
+        """Reset every per-action model to its cold-start state."""
+
+
+class PerActionMetaModel(BaseMetaModel):
```
shaharbar1: Shouldn't PerActionMetaModel and BaseMetaModel for the time being? Or how do you see the inheritance on the next PR?

shaharbar1 (changes requested):
A lot of improvement, but still some leftover fundamental issues

pybandits/actions_manager.py
AnastasiiaKabeshova: Updated

pybandits/cmab.py
AnastasiiaKabeshova: the cast(Dict, ...) is necessary because state["actions_manager"] is typed as Serializable (a recursive union), and mypy can't narrow it for dict subscripting or in tests without help.

pybandits/meta_model.py
AnastasiiaKabeshova: The split is intentional and correct. BaseMetaModel is the abstract interface; PerActionMetaModel is concrete implementation #1. The next PR adds a NeuralLinearMetaModel(BaseMetaModel) as a sibling — no changes to PerActionMetaModel needed.

pybandits/actions_manager.py
```diff
@@ -824,21 +700,14 @@ class SmabActionsManager(ActionsManager, Generic[SmabModelType]):
...
-    actions: Dict[ActionId, SmabModelType]
-
-    @field_validator("actions", mode="after")
-    @classmethod
-    def all_actions_have_same_number_of_objectives(cls, actions: Dict[ActionId, SmabModelType]):
-        n_objs_per_action = [len(beta.models) if hasattr(beta, "models") else None for beta in actions.values()]
-        if len(set(n_objs_per_action)) != 1:
-            raise ValueError("All actions should have the same number of objectives")
-        return actions
+    meta_model: SmabMetaModel[SmabModelType]
```
shaharbar1: It feels a little redundant to use generic types here, why SmabMetaModel needs SmabModelType, isn't it obvious?

pybandits/actions_manager.py
```diff
@@ -931,53 +771,14 @@ class CmabActionsManager(ActionsManager, Generic[CmabModelType]):
...
-        for action in action_models[1:]:
-            test_action = cls._maybe_crawl_model(action)
-            if not cls._get_expected_context_size(first_action) == cls._get_expected_context_size(action):
-                raise AttributeError("All actions should have the same input size.")
-            if not test_first_action.update_method == test_action.update_method:
-                raise AttributeError("All actions should have the same update method.")
-            if not test_first_action.update_kwargs == test_action.update_kwargs:
-                raise AttributeError("All actions should have the same update kwargs.")
-        return v
+    meta_model: CmabMetaModel[CmabModelType]
```
shaharbar1: Same comment as for the smab

pybandits/actions_manager.py
```diff
@@ -101,36 +115,38 @@ class ActionsManager(PyBanditsBaseModel, ABC):
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
-    @field_validator("actions", mode="after")
+    @property
+    def actions(self) -> Dict[ActionId, BaseModel]:
+        """Per-action models, delegated to ``meta_model.actions_view()``."""
+        return self.meta_model.actions_view()
```
shaharbar1: Why is it actions_view and not just actions?

pybandits/actions_manager.py
```diff
@@ -152,13 +168,17 @@ def __init__(
...
         actions_with_change: Optional[Set[Tuple[ActionId, NonNegativeInt]]] = None,
+        meta_model: Optional[BaseMetaModel] = None,
     ):
         kwargs = kwargs or {}
-        actions = self._instantiate_actions(
-            actions=actions, action_ids=action_ids, quantitative_action_ids=quantitative_action_ids, kwargs=kwargs
-        )
+        if meta_model is None:
+            meta_model_cls = type(self)._get_meta_model_cls()
+            actions = meta_model_cls._instantiate_actions(
```
shaharbar1: This is over complicated. Metamodel should handle the actions internally, not inside the actions manager. Just pass all arguments to metamodel and he should take care...

pybandits/actions_manager.py
```diff
@@ -152,13 +168,17 @@ def __init__(
         quantitative_action_ids: Optional[Set[ActionId]] = None,
         kwargs: Optional[Dict[str, Any]] = None,
         actions_with_change: Optional[Set[Tuple[ActionId, NonNegativeInt]]] = None,
+        meta_model: Optional[BaseMetaModel] = None,
```
shaharbar1, line 186: You should validate that you get either of actions or metamodel, and not allow both

pybandits/cmab.py
shaharbar1: So how did it work in the old code before your change?

pybandits/meta_model.py
shaharbar1: That's fine, but than what you'll have basementmodel with two children and each child would have single child. That can be shrunk into basemetamodrl having two children, one for smab and one for cmab

shaharbar1 (changes requested):

pybandits/actions_manager.py
AnastasiiaKabeshova: refactored

pybandits/actions_manager.py
AnastasiiaKabeshova: refactored

pybandits/actions_manager.py
```diff
@@ -152,13 +175,27 @@ def __init__(
...
-        )
+        action_args = (actions, action_ids, quantitative_action_ids)
+        if meta_model is not None and any(a is not None for a in action_args):
+            raise ValueError(
+                "Provide either 'meta_model' or action-construction arguments "
+                "('actions', 'action_ids', 'quantitative_action_ids'), not both."
+            )
+        if meta_model is None:
+            meta_model = (
+                type(self)
```
shaharbar1: why not just self? I don't think you need type()

pybandits/actions_manager.py
```diff
@@ -152,13 +175,27 @@ def __init__(
...
+                ._get_meta_model_cls()
+                .from_kwargs(
```
shaharbar1: I don't like the name from_kwargs, it's ambigious with the fact that it takes both arguments and kwargs

pybandits/actions_manager.py
```diff
@@ -824,28 +717,31 @@ class SmabActionsManager(ActionsManager, Generic[SmabModelType]):
...
 
-    actions: Dict[ActionId, SmabModelType]
+    meta_model: PerActionMetaModel[SmabModelType]  # type: ignore[valid-type]
 
-    @field_validator("actions", mode="after")
-    @classmethod
-    def all_actions_have_same_number_of_objectives(cls, actions: Dict[ActionId, SmabModelType]):
-        n_objs_per_action = [len(beta.models) if hasattr(beta, "models") else None for beta in actions.values()]
+    @model_validator(mode="after")
+    def all_actions_have_same_number_of_objectives(self) -> "SmabActionsManager":
```
shaharbar1, line 738: Thanks for the fix. pydantic2 is finally here.

pybandits/actions_manager.py
```diff
@@ -952,40 +818,34 @@ def _maybe_crawl_model(model: CmabModelType):
...
 
     @validate_call(config=dict(arbitrary_types_allowed=True))
     def update(
         self,
         actions: List[ActionId],
         rewards: Union[List[BinaryReward], List[List[BinaryReward]]],
-        quantities: Optional[List[Union[float, List[float], None]]],
-        context: np.ndarray,
+        quantities: Optional[List[Union[float, List[float], None]]] = None,
+        context: Optional[np.ndarray] = None,
```
shaharbar1: Why is context optional? It's obligatory for cmab.

pybandits/meta_model.py
```diff
@@ -0,0 +1,451 @@
...
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    @property
+    @abstractmethod
+    def action_ids(self) -> List[ActionId]:
+        """Action identifiers covered by this meta-model."""
+
+    @abstractmethod
+    def get_actions(self) -> Dict[ActionId, BaseModel]:
```
shaharbar1: I don't under why do we need this?

pybandits/meta_model.py
```diff
@@ -0,0 +1,451 @@
...
+    """Abstract base for objects that own per-action state and dispatch logic.
+
+    The manager class (e.g. ``CmabActionsManager``) delegates ``sample_proba``,
+    ``update``, and ``reset`` to a meta-model instance. The meta-model owns the
+    per-action models (exposed for backward-compatible read access via
+    ``get_actions``) and decides whether they are dispatched independently or
+    with coordinated shared state.
+    """
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
shaharbar1, line 120: Why isn't actions an attribute of the abstract class?

pybandits/meta_model.py
```diff
@@ -0,0 +1,451 @@
...
+    instances during validation, so the original model references are preserved.
+    This guarantees that mutations via ``update()`` are visible to callers that
+    hold references to the original per-action model objects.
+
+    Domain invariants (e.g. same number of objectives for smab; consistent
+    input size / update method for cmab) are validated on the respective manager
+    subclasses, not here.
+    """
+
+    actions: Dict[ActionId, ActionModelType]  # type: ignore[valid-type]
```
shaharbar1: I think it should already be on the base class

tests/test_actions_manager.py
```diff
@@ -294,86 +300,86 @@ def cold_start(cls):
...
+    mocker.patch("pybandits.meta_model.issubclass", return_value=True)
     kwargs = {"irrelevant_param": 1}
     kwargs_copy = kwargs.copy()
     (
         _,
         _,
         action_general_kwargs,
         quantitative_action_general_kwargs,
-    ) = ActionsManager._extract_action_model_class_and_attributes(kwargs_copy)
+    ) = PerActionMetaModel._extract_action_model_class_and_attributes(kwargs_copy)
```
shaharbar1: All these tests should move to test_meta_model as they are not testing actions manager functionality anymore

tests/test_actions_manager.py
```diff
@@ -1095,3 +1068,96 @@ def test_update_adaptive_window_size_validation(
...
+def test_manager_update_delegates_to_meta_model(monkeypatch, n_features=3):
+    actions = {
+        "a1": BayesianNeuralNetwork.cold_start(n_features=n_features),
+        "a2": BayesianNeuralNetwork.cold_start(n_features=n_features),
+    }
+    manager = CmabActionsManager[BayesianNeuralNetwork](actions=actions)
+    captured = []
+    monkeypatch.setattr(PerActionMetaModel, "update", lambda self, **kwargs: captured.append(kwargs), raising=True)
+
+    context = np.array([[0.1, 0.2, 0.3], [0.4, 0.5, 0.6]])
```
shaharbar1: Please refrain from hard coded constants all along the new tests. Try to use parameterize and/or hypothesis and/or pytest fixtures to make the tests more generic

shaharbar1 (changes requested):

pybandits/actions_manager.py
```diff
@@ -86,51 +110,50 @@ class ActionsManager(PyBanditsBaseModel, ABC):
...
         The confidence level for the adaptive window. None for skipping the change point detection.
     """
 
-    actions: Dict[ActionId, BaseModel]
+    meta_model: BaseMetaModel
     delta: Optional[PositiveProbability] = None
-    _no_change_point: ClassVar[NonNegativeInt] = -1
-    _min_adaptive_window_size: ClassVar[NonPositiveInt] = 10000
+    _no_change_point: ClassVar[int] = -1
+    _min_adaptive_window_size: ClassVar[int] = 10000
```
shaharbar1: It should stay NonPositiveInt, or become PositiveInt

pybandits/actions_manager.py
```diff
@@ -952,32 +817,26 @@ def _maybe_crawl_model(model: CmabModelType):
...
-        else:
-            raise TypeError(f"Unsupported model type: {type(model)}")
+        return model.input_dim
 
-    @field_validator("actions", mode="after")
-    @classmethod
-    def check_models(cls, v):
-        action_models = list(v.values())
+    @model_validator(mode="after")
+    def check_models(self) -> "CmabActionsManager":
```
shaharbar1: Shouldn't this move to CmabMetaModel?

pybandits/actions_manager.py
```diff
@@ -952,32 +817,26 @@ def _maybe_crawl_model(model: CmabModelType):
...
         first_action = action_models[0]
-        test_first_action = cls._maybe_crawl_model(first_action)
+        test_first_action = self._maybe_crawl_model(first_action)
         for action in action_models[1:]:
-            test_action = cls._maybe_crawl_model(action)
-            if not cls._get_expected_context_size(first_action) == cls._get_expected_context_size(action):
-                raise AttributeError("All actions should have the same input size.")
+            test_action = self._maybe_crawl_model(action)
+            if not self._get_expected_context_size(first_action) == self._get_expected_context_size(action):
+                raise ValueError("All actions should have the same input size.")
```
shaharbar1: Why did you replace AttributeError(s) with ValueError(s)?
This is already an instantiated model so these are model attributes...

pybandits/meta_model.py
```diff
@@ -0,0 +1,454 @@
...
+    @classmethod
+    def at_least_one_action_is_defined(cls, v: Dict[ActionId, ActionModelType]) -> Dict[ActionId, ActionModelType]:
+        if len(v) == 0:
+            raise ValueError("At least one action should be defined.")
+        elif len(v) == 1:
+            warnings.warn("Only a single action was supplied. This MAB will be deterministic.")
+        return v
+
+    @classmethod
+    def build(
```
shaharbar1: Why isn't it an init function like in the actionsmanager or mab, that take in init multiple arguments and handle them to feed pydantic BaseModel with the actual model attributes?

shaharbar1: I mean, you can do it with optional parameters, like we did in actionsmanager init. How do you see it? Maybe I'm biased :-)

pybandits/meta_model.py
```diff
@@ -0,0 +1,454 @@
...
+        )
+        return cls(actions=actions)  # type: ignore[call-arg]
+
+    @classproperty
+    def _action_model_classes(cls) -> Tuple[Type[BaseModel], ...]:
+        """Extract action model classes from the ``actions`` field annotation (``Dict[ActionId, T]`` → ``T``)."""
+        actions_annotation = cls.model_fields["actions"].annotation
+        type_args = get_args(actions_annotation)  # (ActionId, T)
+        if not type_args or len(type_args) < 2:
+            return (BaseModel,)
```
shaharbar1: Does these returns of (BaseModel, ) make sense? BaseModel is anyhow abstract, why would we ever return it? shouldn't you raise an error here?

pybandits/meta_model.py
```diff
@@ -0,0 +1,454 @@
...
+                [model_cold_start, quantitative_model_cold_start],
+                [action_general_kwargs, quantitative_action_general_kwargs],
+                [action_specific_kwargs, quantitative_action_specific_kwargs],
+            ):
+                for a in _action_ids:
+                    actions[a] = cold_start(**general_kwargs, **specific_kwargs.get(a, {}))
+
+        return actions
+
+    def sample_proba(
```
shaharbar1, line 377: Consider making this in the abstract BaseModel with some abstract _preprocessing method. Anyhow (I mean either for smab or NLB), you shall call predict for all actions, in NLB you'll just have a joint preprocessing, and in smab it will be identity mapping.
So eventually, this piece of code is relevant for all models, either PerActionsMetaModel or not

tests/test_actions_manager.py
```diff
@@ -899,13 +758,13 @@ def mock_change_detection(action_id, *args, **kwargs):
...
         ),
-        ({"action1": Beta(n_successes=7, n_failures=3)}, 8),
+        ({"action1": Beta(n_successes=7, n_failures=3), "action2": Beta()}, 8),
     ],
 )
 def test_get_expected_memory_length_with_base_model_so(actions, expected_len):
     """Test _get_expected_memory_length with BaseModelSO models."""
-    actual_length = ActionsManager._get_expected_memory_length(actions)
-    assert actual_length == expected_len
+    manager = SmabActionsManagerSO(actions=actions)
```
shaharbar1: Why don't you keep using the class method of the abstract class?

shaharbar1: If you do this, you also need to test for cmab...
I suggest you keep the original method a classmethod as it is fundumental, or replace here with DummyActionsManager

tests/test_actions_manager.py
```diff
@@ -1095,3 +921,129 @@ def test_update_adaptive_window_size_validation(
...
+def test_cmab_manager_update_delegates_to_meta_model(monkeypatch, n_features, batch_size):
+    action_ids = ("a1", "a2")
+    _, manager = _build_cmab_manager(action_ids, n_features)
+    captured = []
+    monkeypatch.setattr(PerActionMetaModel, "update", lambda self, **kwargs: captured.append(kwargs), raising=True)
+
+    rng = np.random.default_rng(0)
+    sampled_actions = [action_ids[i % len(action_ids)] for i in range(batch_size)]
+    rewards = rng.integers(0, 2, size=batch_size).tolist()
+    context = rng.normal(size=(batch_size, n_features))
```
shaharbar1: I think we have generate_context and generate_rewards or something like that under test utils.

tests/test_actions_manager.py
```diff
@@ -1095,3 +921,129 @@ def test_update_adaptive_window_size_validation(
...
+    manager.update(actions=sampled_actions, rewards=rewards, context=context, quantities=None)
+
+    assert len(captured) == 1
+    assert captured[0]["actions"] == sampled_actions
+    assert captured[0]["rewards"] == rewards
+    assert captured[0]["quantities"] is None
+    np.testing.assert_array_equal(captured[0]["context"], context)
+
+
+@pytest.mark.parametrize("batch_size", [1, 3, 10])
```
shaharbar1: hypothesis given feels more natural here

tests/test_actions_manager.py
```diff
@@ -1095,3 +921,129 @@ def test_update_adaptive_window_size_validation(
...
+    assert set(result.keys()) == valid_ids
+    assert captured["rng"] is rng
+    assert captured["valid_action_ids"] == valid_ids
+    np.testing.assert_array_equal(captured["kwargs"]["context"], context)
+
+
+@pytest.mark.parametrize("valid_ids", [{"a1"}, {"a1", "a2"}])
+@pytest.mark.parametrize("n_samples", [1, 5])
+def test_smab_manager_sample_proba_delegates_to_meta_model(monkeypatch, valid_ids, n_samples):
+    action_ids = ("a1", "a2", "a3")
```
shaharbar1: make this an argument of the test

tests/test_meta_model.py
```diff
@@ -0,0 +1,339 @@
...
+    assert meta.actions["a1"] is beta1
+    assert meta.actions["a2"] is beta2
+
+
+def test_sample_proba_dispatches_to_each_action():
+    actions = {"a1": Beta(), "a2": Beta(), "a3": Beta()}
+    meta = PerActionMetaModel(actions=actions)
+    rng = np.random.default_rng(42)
+    # Beta.sample_proba requires n_samples; the meta-model forwards kwargs.
+    result = meta.sample_proba(rng=rng, n_samples=3)
```
shaharbar1: hard coded constants...

tests/test_meta_model.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+def test_update_with_quantities_only_passes_quantities_when_any_present(monkeypatch):
+    """Mirrors the historical cmab behaviour: when all quantities for an action
+    are falsy (`any(...)` is False), the per-action update is called without
+    `quantities`; when any are truthy, the full list is passed through.
+
+    We capture call kwargs by overriding ``BaseBeta._update`` at the class
+    level via monkeypatch (so it accepts arbitrary kwargs)."""
+    import pybandits.model as model_mod
```
shaharbar1: No local imports please

tests/test_meta_model.py
```diff
@@ -0,0 +1,339 @@
...
+    meta.reset()
+    assert actions["a1"].n_successes == 1
+    assert actions["a2"].n_successes == 1
+
+
+def test_update_with_only_some_actions_in_batch_leaves_others_untouched():
+    a1, a2 = Beta(), Beta()
+    meta = PerActionMetaModel(actions={"a1": a1, "a2": a2})
+    a2_state_before = (a2.n_successes, a2.n_failures)
+    meta.update(actions=["a1"] * 5, rewards=[1] * 5)
```
shaharbar1: move 5 to a test argument

shaharbar1 (changes requested):

tests/test_meta_model.py
```diff
@@ -0,0 +1,401 @@
...
+    actions = make_beta_actions(n)
+    meta = PerActionMetaModel(actions=actions)
+    for a, beta in actions.items():
+        assert meta.actions[a] is beta
+
+
+@pytest.mark.parametrize("n_actions, n_samples", [(2, 1), (3, 3), (5, 7)])
+def test_sample_proba_dispatches_to_each_action(make_beta_meta, n_actions, n_samples):
+    meta = make_beta_meta(n_actions)
+    rng = np.random.default_rng(42)
```
shaharbar1: There is an rng fixture defined under conftest, use it.

tests/test_meta_model.py
```diff
@@ -0,0 +1,401 @@
...
+    result = meta.sample_proba(rng=rng, n_samples=n_samples)
+    assert set(result.keys()) == set(meta.action_ids)
+    assert all(len(v) == n_samples for v in result.values())
+
+
+@pytest.mark.parametrize("valid_ratio", [0.5, 1.0])  # half vs all of the action set
+def test_sample_proba_restricts_to_valid_action_ids(make_beta_meta, valid_ratio):
+    meta = make_beta_meta(4)
```
shaharbar1: 4 -> test argument

tests/test_meta_model.py
```diff
@@ -0,0 +1,401 @@
...
+    assert all(meta.actions[a].n_successes == 1 for a in action_ids)
+    assert all(meta.actions[a].n_failures == 1 for a in action_ids)
+
+
+def test_update_with_only_some_actions_in_batch_leaves_others_untouched(make_beta_actions):
+    actions = make_beta_actions(2)
+    a0, a1 = actions["a0"], actions["a1"]
+    meta = PerActionMetaModel(actions=actions)
+    a1_state_before = (a1.n_successes, a1.n_failures)
+    meta.update(actions=["a0"] * 5, rewards=[1] * 5)
```
shaharbar1: 5 -> test argument

tests/test_meta_model.py
```diff
@@ -0,0 +1,401 @@
...
+    action_ids = list(meta.action_ids)
+    meta.update(actions=action_ids, rewards=[1] * n)
+    assert all(meta.actions[a].n_successes > 1 for a in action_ids)
+    meta.reset()
+    assert all(meta.actions[a].n_successes == 1 for a in action_ids)
+    assert all(meta.actions[a].n_failures == 1 for a in action_ids)
+
+
+def test_update_with_only_some_actions_in_batch_leaves_others_untouched(make_beta_actions):
+    actions = make_beta_actions(2)
```
shaharbar1: 2 -> test argument

tests/test_meta_model.py
```diff
@@ -0,0 +1,401 @@
...
+    meta = make_beta_meta(2)
+    action_list = list(meta.action_ids)  # one row per action
+    context = np.arange((len(action_list) + context_extra_rows) * n_features, dtype=float).reshape(-1, n_features)
+    meta.update(actions=action_list, rewards=[1] * len(action_list), context=context)
+    # Each per-action update should see exactly its own row (1 row, n_features cols).
+    assert all(call["context"].shape == (1, n_features) for call in spy_beta_update)
+
+
+def test_update_context_length_guard_rejects_short_context(make_beta_meta):
+    meta = make_beta_meta(2)
```
shaharbar1: 2 -> test argument

tests/utils.py
```diff
@@ -38,6 +38,15 @@ def sample_with_replacement(source: list, length: PositiveInt):
     return [random.choice(source) for _ in range(length)]
 
 
+def make_action_ids(n: int) -> Tuple[str, ...]:
```
shaharbar1, line 41: I like it!

shaharbar1 (approved):

PR #148, opened 2026-06-15
Add Neural-Linear contextual bandit (shared backbone + per-arm Bayesi…

AnastasiiaKabeshova:
### Changes:

* Add pybandits/neural_linear.py — Neural-Linear cMAB built on the meta-model abstraction
  - MLPBackbone: shared deterministic encoder, Glorot init, numpy embed() reusing forward_layers
  - BayesianLinearHead (+ BayesianLinearHeadCC): per-arm logistic head over a NormalArray posterior
  - BackboneHeadsMetaModel(BaseCmabMetaModel): joint SVI per update() — backbone via numpyro.param, heads via numpyro.sample; streaming prior = previous posterior; kl_weight rides the existing kl_annealing_factor path; no replay buffer
  - action_ids→heads cold-start done in a model_validator(mode="before") (not __init__) so the meta_model Union validates from a dict without a raw TypeError
* Add pybandits/model/bnn/_svi.py — extract forward_layers + run_svi_epochs as backend-agnostic helpers shared by the BNN and the backbone
* Refactor pybandits/model/bnn/network.py — _forward_layers and _run_svi_training_loop now delegate to _svi.py (behaviour-preserving); drop now-unused tqdm/loguru imports
* Add BaseCmabMetaModel to pybandits/meta_model.py — common abstract parent of CmabPerActionMetaModel and BackboneHeadsMetaModel
* Update pybandits/actions_manager.py — widen CmabActionsManager.meta_model to Union[CmabPerActionMetaModel[CmabModelType], BackboneHeadsMetaModel]
  - override _get_meta_model_cls() to keep the default actions=/action_ids= path on the per-action meta-model
  - Fix _validate_update_params() standard-action check: isinstance (Model, ModelMO) → not isinstance QuantitativeModel, so non-Model heads pass
* Add CmabBernoulli.cold_start() to pybandits/cmab.py — backbone_hidden_dims selects Neural-Linear; otherwise delegates to BaseMab.cold_start so the per-action path is byte-identical
  - backbone activation is backbone_activation= so activation= stays free to reach the per-action BNN via **bnn_kwargs
  - input_dim property returns backbone.n_features for the Neural-Linear path
* Fix pybandits/mab.py default_action validator — standard (str) default must not map to a QuantitativeModel (was an isinstance (Model, ModelMO) check that rejected heads)
  - widen _select_epsilon_greedy_action actions param BaseModel → PyBanditsBaseModel
* Update pybandits/strategy/base.py — widen SingleObjectiveStrategy.select_action actions param BaseModel → PyBanditsBaseModel
* Add tests/test_neural_linear.py — backbone shape/determinism, head probability range, meta-model action_ids/sample_proba/update/reset, absent-arm carry-forward
* Add CmabBernoulli.cold_start tests to tests/test_cmab.py — Neural-Linear round-trip (predict→update) and per-action fallback regression guard

## Summary by CodeRabbit

* **New Features**
  * Added `cold_start` for contextual bandit Bernoulli models, including a shared backbone option for initialization and prediction/update.
  * Introduced a unified contextual-bandit meta-model enabling joint variational training across actions (with optional shared backbone).
* **Bug Fixes**
  * Strengthened validation for action quantities and shared-backbone adaptive-window settings; disallowed incompatible configurations.
  * `input_dim` now derives from the shared backbone when used.
* **Tests**
  * Expanded unit tests for shared-backbone cold-start, unified meta-model updates/reset behavior, and meta-model wiring across managers.


pybandits/strategy/base.py
```diff
@@ -88,7 +88,7 @@ class SingleObjectiveStrategy(BaseStrategy, ABC):
     def select_action(
         self,
         p: Dict[ActionId, Union[float, Callable[[np.ndarray], float]]],
-        actions: Dict[ActionId, BaseModel],
+        actions: Dict[ActionId, PyBanditsBaseModel],
```
shaharbar1: Why?

pybandits/actions_manager.py
```diff
@@ -79,8 +79,8 @@
     BetaMO,
     BetaMOCC,
     Model,
-    ModelMO,
 )
+from pybandits.neural_linear import BackboneHeadsMetaModel
```
shaharbar1: I don't like the name neural_linear. What if we have 2 BNN layers on top of the joint backbone?
I suggest something like neural_backbone

pybandits/neural_linear.py
```diff
@@ -0,0 +1,447 @@
...
+    @staticmethod
+    def _init_params(
+        n_features: int, hidden_dims: List[int], embedding_dim: int, random_seed: Optional[int]
+    ) -> Dict[str, List[np.ndarray]]:
+        """Glorot-uniform weights and zero biases for each layer; deterministic given ``random_seed``."""
+        rng = np.random.default_rng(random_seed)
+        dims = [n_features, *hidden_dims, embedding_dim]
+        weights, biases = [], []
+        for input_dim, output_dim in zip(dims[:-1], dims[1:]):
+            limit = np.sqrt(6.0 / (input_dim + output_dim))
```
shaharbar1: Put 6 in constant.
The init should be activation dependent: He for relu/gelu, Xavier for the tanh/sigmoid. Please add references

pybandits/neural_linear.py
```diff
@@ -0,0 +1,447 @@
...
+    """Cost-control variant of :class:`BayesianLinearHead` (mirrors ``BetaCC``).
+
+    Defined for symmetry with the BNN cost-control models; not wired into a bandit yet (the
+    single-objective Neural-Linear path is the only one exposed in this phase).
+    """
+
+    cost: NonNegativeFloat
+
+
+class BackboneHeadsMetaModel(BaseCmabMetaModel):
```
shaharbar1: Here there is an architectural choice that needs to be examined carefully.
IMHO, you can have a single MetaModel for cmab. The backbone should be optional. The methodology should be of this metamodel. I don't see why you need the other meta model. WDYT?

pybandits/neural_linear.py
```diff
@@ -0,0 +1,447 @@
...
+        )
+
+    def update_params(self, params: Dict[str, Any]) -> "MLPBackbone":
+        """Return a copy with ``_params`` replaced (Pydantic immutability)."""
+        updated = self.model_copy()
+        updated._params = params
+        return updated
+
+
+class BayesianLinearHead(PyBanditsBaseModel):
```
shaharbar1: I don't think you need to restrict to linear head. What if I want DNN -> 2 BNN layers?

pybandits/neural_linear.py
shaharbar1: If you accepy my offer, then BayesianLinearHead becomes redundant. You can just use the BNN class as is and remove the update method from it.

pybandits/neural_linear.py
```diff
@@ -0,0 +1,447 @@
...
+            return jax.lax.scan(_body, state, epoch_factors)
+
+        final_state, _ = run_svi_epochs(
+            _run_one_epoch, svi_state, [factors], restore_best=True, desc="neural-linear SVI"
+        )
+        self._store_results(svi.get_params(final_state), site_names)
+
+    def _store_results(self, params: Dict[str, Any], site_names: Dict[ActionId, str]) -> None:
+        """Write back the trained backbone point estimates and per-arm head posteriors."""
+        new_weights = [np.asarray(params[f"bb_W_{i}"]) for i in range(len(self.backbone._params["W"]))]
```
shaharbar1: don't use the hard coded strings, have them as class var with dedicated function to get the name by layer index, like we do on the BNN class.

shaharbar1 (changes requested):

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,227 @@
...
+}
+_JAX_ACTIVATIONS: Dict[str, Any] = {
+    "tanh": jax.nn.tanh,
+    "relu": jax.nn.relu,
+    "sigmoid": jax.nn.sigmoid,
+    "gelu": lambda x: jax.nn.gelu(x, approximate=False),
+}
+
+
+def _init_limit(input_dim: int, output_dim: int, activation: ActivationFunctions) -> float:
```
shaharbar1: I think it should be a method under the class. Not sure if static or not, depends on whether you can read the activation from self.

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,227 @@
...
+from pybandits.model.bnn._svi import forward_layers
+from pybandits.model.bnn._typing import (
+    ActivationFunctions,
+    _numpy_gelu,
+    _numpy_relu,
+    _numpy_sigmoid,
+)
+
+# Uniform-init half-width numerator: limit = sqrt(NUMERATOR / fan), shared by He and Glorot/Xavier.
+_INIT_LIMIT_NUMERATOR = 6.0
```
shaharbar1: All these constants should become class vars under BaseBackbone

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,227 @@
...
+    * **Xavier/Glorot** uniform — ``sqrt(6 / (fan_in + fan_out))`` — for ``tanh``/``sigmoid``
+      (Glorot & Bengio 2010, *Understanding the difficulty of training deep feedforward networks*,
+      https://proceedings.mlr.press/v9/glorot10a.html).
+    """
+    if activation in _HE_ACTIVATIONS:
+        return float(np.sqrt(_INIT_LIMIT_NUMERATOR / input_dim))
+    return float(np.sqrt(_INIT_LIMIT_NUMERATOR / (input_dim + output_dim)))
+
+
+class BaseBackbone(PyBanditsBaseModel, ABC):
```
shaharbar1: Why do you need inheritance here? Why not have just MLPBackbone? It's just a single child.

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,227 @@
...
+# SOFTWARE.
+"""Shared deterministic DNN backbone (preprocessor) for the contextual-bandit meta-model.
+
+The backbone maps a context matrix ``(n, n_features)`` to an embedding ``(n, embedding_dim)`` with
+point-estimate (deterministic) weights. It is shared across arms and trained jointly with the per-arm
+Bayesian heads inside one SVI pass (its weights enter the joint NumPyro model as ``numpyro.param``).
+Pydantic immutability is preserved by returning a new instance from :meth:`update_params`.
+"""
+
+from abc import ABC, abstractmethod
```
shaharbar1: Consider creating a BaseDNN with two children: MLPBackbone and the BaseBNN, because a lot of things are joint, e.g. the activations' dict.

pybandits/model/bnn/network.py
```diff
@@ -313,9 +312,12 @@ def validate_bias_calibrated(self) -> "BaseBayesianNeuralNetwork":
...
     @classmethod
-    def get_embedding_var_name(cls, feat_index: int) -> str:
-        """Return the NumPyro variable name for a categorical embedding matrix."""
-        return f"{cls._embedding_var_name}_{feat_index}"
+    def get_embedding_var_name(cls, feat_index: int, prefix: str = "") -> str:
+        """Return the NumPyro variable name for a categorical embedding matrix.
+
+        ``prefix`` namespaces the site for joint multi-arm models; ``""`` keeps the standalone name.
+        """
+        return f"{prefix}{cls._embedding_var_name}_{feat_index}"
```
shaharbar1: Why do we need this prefix?

shaharbar1: Can we make it a sort of classvar or classproperty?

pybandits/cmab.py
```diff
@@ -301,6 +307,119 @@ class CmabBernoulli(BaseCmabBernoulli):
     strategy: ClassicBandit
     _predict_with_proba: bool = False
 
+    @classmethod
+    def cold_start(
```
shaharbar1: You need to convince me that this is needed and not covered by the dynamic cold start under the mab class.

tests/test_cmab.py
```diff
@@ -1249,3 +1251,91 @@ def test_random_seed_propagates_to_bnn(
...
+)
+def test_cmab_bernoulli_cold_start_rejects_cross_path_arguments(
+    kwargs: dict, expected_exception: type, match: str
+) -> None:
+    """Arguments belonging to the other model path raise instead of being silently ignored."""
+    with pytest.raises(expected_exception, match=match):
+        CmabBernoulli.cold_start(action_ids={"a", "b"}, n_features=5, **kwargs)
+
+
+def test_cmab_bernoulli_nlb_input_dim() -> None:
```
shaharbar1: Please use the skill for writing tests.

shaharbar1 (changes requested):

pybandits/model/bnn/network.py
AnastasiiaKabeshova: Replaced with numpyro.handlers.scope

pybandits/cmab.py
AnastasiiaKabeshova: backbone is now built from kwargs in CmabMetaModel

tests/test_cmab.py
AnastasiiaKabeshova: done

pybandits/model/bnn/backbone.py
AnastasiiaKabeshova: Added new _dnn.py to avoid this inheritance

pybandits/model/bnn/_dnn.py
```diff
@@ -0,0 +1,73 @@
...
+"""
+
+from typing import TYPE_CHECKING, Any, ClassVar, Dict
+
+import jax
+import numpy as np
+
+from pybandits.model.bnn._typing import _numpy_gelu, _numpy_relu, _numpy_sigmoid
+
+if TYPE_CHECKING:
```
shaharbar1: Why do we need this if TYPE_CHECKING here and later?

pybandits/model/bnn/_dnn.py
```diff
@@ -0,0 +1,73 @@
...
+    from pybandits.model.bnn._typing import ActivationFunctions
+
+
+class BaseDNN:
```
shaharbar1: Base is for abstract classes usually. Maybe make it DNNMixin or FFNMixin?

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,188 @@
...
+    activated; the final layer is a linear projection to the embedding space (the per-arm Bayesian
+    heads supply the decision non-linearity). Point-estimate weights/biases are stored as plain lists
+    (``weights``/``biases``) so the model serialises via ``get_state``/``from_state``; numpy views
+    (``_W``/``_b``) are rebuilt on init for the forward pass.
+    """
+
+    # Uniform-init half-width numerator: limit = sqrt(NUMERATOR / fan), shared by He and Glorot/Xavier.
+    _INIT_LIMIT_NUMERATOR: ClassVar[float] = 6.0
+    # Activations whose negative-half saturation calls for He init (variance ~ 2/fan_in).
+    _HE_ACTIVATIONS: ClassVar[Tuple[str, ...]] = ("relu", "gelu")
```
shaharbar1: No need for capital letters once it's class vars

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def model_post_init(self, __context: Any) -> None:
+        """Rebuild numpy weight/bias views from the serialisable list fields."""
+        self._W = [np.asarray(w, dtype=float) for w in self.weights]
+        self._b = [np.asarray(b, dtype=float) for b in self.biases]
+
+    @property
+    def _params(self) -> Dict[str, List[np.ndarray]]:
+        """Numpy weight/bias views keyed ``"W"`` / ``"b"`` (consumed by the joint SVI engine)."""
+        return {"W": self._W, "b": self._b}
```
shaharbar1: Put "W" and "b" in a class var or use the already existing class vars...

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,188 @@
...
+            embedding_dim=embedding_dim,
+            hidden_dims=hidden_dims,
+            activation=activation,
+            random_seed=random_seed,
+            weights=[w.tolist() for w in params["W"]],
+            biases=[b.tolist() for b in params["b"]],
+        )
+
+    def embed(self, x: np.ndarray) -> np.ndarray:
+        """Deterministic forward pass producing the shared embedding ``(n, embedding_dim)``."""
```
shaharbar1: add Parameters and return sections to docstring all over the place

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,479 @@
...
+    """
+
+    actions: Dict[ActionId, CmabHeadType]  # type: ignore[assignment]
+    backbone: Optional[MLPBackbone] = None
+    random_seed: Optional[NonNegativeInt] = None
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    # Backbone output width used when ``backbone_hidden_dims`` is requested without an ``embedding_dim``.
+    _DEFAULT_EMBEDDING_DIM: ClassVar[int] = 32
```
shaharbar1: Why not use cardinality //4 as default like we did for the BNN embeddings?

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,479 @@
...
+        self._rng_key = jax.random.PRNGKey(self.random_seed if self.random_seed is not None else 0)
+
+    @model_validator(mode="after")
+    def _check_models(self) -> "CmabMetaModel":
+        """Cross-arm consistency: every head must share input size / update method / update kwargs.
+
+        Adapted from the old ``CmabPerActionMetaModel.check_models``; the heads' input dim must also
+        match the backbone's ``embedding_dim`` when a backbone is present.
+        """
+        if not self.actions:
```
shaharbar1: This code replicates other code snippets. Can't it be put to the father class?

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,479 @@
...
+            )
+        return self
+
+    # --------------------------------------------------------------------- head-shape helpers
+    # A "head" is one of: a single BNN (SO/CC/DP), an MO head (a list of per-objective BNNs), or a
+    # quantitative head (wraps a `.bnn` whose input is [quantity ‖ context]). These helpers normalise
+    # the three shapes for the joint engine.
+    @staticmethod
+    def _rep_bnn(head: BaseModel) -> BaseBayesianNeuralNetwork:
+        """A representative underlying BNN (for update_method / update_kwargs / optimizer / eps)."""
```
shaharbar1: Repetitive code...

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,479 @@
...
+        context: Optional[np.ndarray] = None,
+        **kwargs: Any,
+    ) -> Dict[ActionId, SampleProbaResult]:
+        """Embed the context once (if a backbone is set), then sample each (valid) head on it."""
+        if context is None:
+            raise AttributeError("context is required for cmab sample_proba.")
+        z = (
+            self.backbone.embed(context)
+            if self.backbone is not None
+            else np.atleast_2d(np.asarray(context, dtype=float))
```
shaharbar1: context is already np array, why do we need asarray? in addition, hasn't it gone through atleast_2d in the actions manager?

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,479 @@
...
+        **kwargs: Any,
+    ) -> None:
+        """Run one joint SVI pass over the batch (VI-only), then drop it (no replay buffer)."""
+        # VI-only is a property of the (consistency-validated) heads — reject MCMC up front, before
+        # touching any per-arm state, so a misconfigured meta-model fails with a clear message.
+        if self._rep_bnn(next(iter(self.actions.values()))).update_method != "VI":
+            raise ValueError("Contextual-bandit joint training is VI-only; set update_method='VI'.")
+        if context is None:
+            raise AttributeError("context is required for cmab update.")
+        _context = np.atleast_2d(np.asarray(context, dtype=float))
```
shaharbar1: Same here, I elieve these transformations were done on the actionsmanager

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,479 @@
...
+        for action_id in set(actions):
+            head = self.actions[action_id]
+            if isinstance(head, BaseBayesianNeuralNetworkMO):
+                n_obj = len(head.models)
+                if rewards_arr.ndim != 2 or rewards_arr.shape[1] != n_obj:
+                    raise AttributeError(
+                        f"multi-objective action '{action_id}' expects rewards with {n_obj} objectives per row."
+                    )
+        if quantities is not None:
+            quantities = quantities[-len(actions) :]
```
shaharbar1: I feel a lot of code is already happening on actions manager. Here for example, I would raise an error, not slice. I believe the actionsmanager does that already.

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,479 @@
...
+        for row_index, action_id in enumerate(actions):
+            arm_to_rows[action_id].append(row_index)
+
+        self._joint_svi_update(_context, arm_to_rows, rewards_arr, quantities)
+        self._increment_counters(arm_to_rows, rewards_arr)
+
+    def reset(self) -> None:
+        """Cold-start the backbone (same seed) and reset every head."""
+        if self.backbone is not None:
+            self.backbone = MLPBackbone.cold_start(
```
shaharbar1: what about self.backbone.reset()?

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,479 @@
...
+            for bnn, obj_index, _q in self._arm_units(self.actions[arm]):
+                # Re-key each unit's namespace-free init arrays under its scope prefix to match the
+                # scoped site names the guide will create.
+                scope = f"{self._unit_scope(arm, obj_index)}{_SCOPE_DIVIDER}"
+                v, s, mu, sg = bnn._collect_guide_init_arrays()
+                values.update({f"{scope}{name}": val for name, val in v.items()})
+                site_sigmas.update({f"{scope}{name}": val for name, val in s.items()})
+                all_mus += mu
+                all_sigmas += sg
+        init_loc_fn, init_scale_fn = BaseBayesianNeuralNetwork._build_guide_init_fns(
```
shaharbar1: change to unprotected method

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,479 @@
...
+
+        # Full-batch joint pass: one epoch of num_steps (per-arm minibatching is a later refinement).
+        num_steps = int(rep._update_kwargs["num_steps"])
+        epoch_factor_arrays = rep._build_kl_annealing_factors([num_steps])
+
+        self._rng_key, subkey = jax.random.split(self._rng_key)
+        svi_state = svi.init(subkey, x, arm_data, 1.0)
+
+        @jax.jit
+        def _run_one_epoch(state: Any, epoch_factors: jax.Array):
```
shaharbar1: Isn't this code already happening on BNN code?
we should not repeat code this way.

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,479 @@
...
+    def _store_backbone(self, params: dict, n_bb_layers: int) -> None:
+        """Write back the trained backbone point estimates (no-op when no backbone)."""
+        if self.backbone is None:
+            return
+        new_w = [np.asarray(params[self.backbone.get_layer_params_name(i)[0]]) for i in range(n_bb_layers)]
+        new_b = [np.asarray(params[self.backbone.get_layer_params_name(i)[1]]) for i in range(n_bb_layers)]
+        self.backbone = self.backbone.update_params({"W": new_w, "b": new_b})
+
+    def _increment_counters(self, arm_to_rows: Dict[ActionId, List[int]], rewards_arr: np.ndarray) -> None:
+        """Mirror each head's ``update`` success/failure bookkeeping (the joint engine bypasses it).
```
shaharbar1: Why do we need this? why bypass the original mechanism?
another code duplication...

shaharbar1 (changes requested):

pybandits/model/bnn/_dnn.py
```diff
@@ -0,0 +1,69 @@
...
+
+from pybandits.model.bnn._typing import ActivationFunctions, _numpy_gelu, _numpy_relu, _numpy_sigmoid
+
+
+class DNNMixin:
+    """Mixin providing the shared activation maps + resolvers for DNN components."""
+
+    # Concrete subclasses (MLPBackbone, BaseBayesianNeuralNetwork) declare this as a Pydantic field;
+    # annotated here (not assigned) only so the resolver properties type-check against the mixin.
+    activation: ActivationFunctions
```
shaharbar1, line 45: You should make it  inherit PyBanditsBaseModel to allow this sytax of attributes properly.

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,314 @@
...
+        Tuple[str, str]
+            The ``(weight_name, bias_name)`` site names for the layer.
+        """
+        return f"{cls.weight_var_name}_{layer_ind}", f"{cls.bias_var_name}_{layer_ind}"
+
+    @classmethod
+    def _init_limit(cls, input_dim: int, output_dim: int, activation: ActivationFunctions) -> float:
+        """Uniform-init half-width for a layer, chosen by activation.
+
+        * **He** uniform — ``sqrt(6 / fan_in)`` — for ``relu``/``gelu`` (He et al. 2015,
```
shaharbar1: Add References header and adopt the references style of the rest of the repo.

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,314 @@
...
+          *Delving Deep into Rectifiers*, https://arxiv.org/abs/1502.01852).
+        * **Xavier/Glorot** uniform — ``sqrt(6 / (fan_in + fan_out))`` — for ``tanh``/``sigmoid``
+          (Glorot & Bengio 2010, *Understanding the difficulty of training deep feedforward networks*,
+          https://proceedings.mlr.press/v9/glorot10a.html).
+
+        Parameters
+        ----------
+        input_dim : int
```
shaharbar1: PositiveInt?

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,314 @@
...
+            The uniform-init half-width ``limit`` (weights drawn from ``U(-limit, limit)``).
+        """
+        if activation in cls._he_activations:
+            return float(np.sqrt(cls._init_limit_numerator / input_dim))
+        return float(np.sqrt(cls._init_limit_numerator / (input_dim + output_dim)))
+
+    @classmethod
+    def _init_params(
+        cls,
+        n_features: int,
```
shaharbar1: PositiveInt?

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,314 @@
...
+            n_features=n_features,
+            embedding_dim=embedding_dim,
+            hidden_dims=hidden_dims,
+            activation=activation,
+            random_seed=random_seed,
+            weights=[w.tolist() for w in weights],
+            biases=[b.tolist() for b in biases],
+        )
+
+    def reset(self) -> "MLPBackbone":
```
shaharbar1: Return Self, import it from typing_extensions

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,314 @@
...
+        ----------
+        x : np.ndarray of shape (n, n_features)
+            Context matrix.
+
+        Returns
+        -------
+        np.ndarray of shape (n, embedding_dim)
+            The shared embedding.
+        """
+        _x = np.atleast_2d(np.asarray(x, dtype=float))
```
shaharbar1: Is this proceesing np.atleast_2d(np.asarray(...)) really needed? Doesn't it happend before? The allocation of np array all over again is just big impedance on speed if not needed.

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,314 @@
...
+        weights_biases : List[Tuple[jnp.ndarray, jnp.ndarray]]
+            Per-layer ``(weight, bias)`` arrays (the ``numpyro.param`` point estimates).
+
+        Returns
+        -------
+        jnp.ndarray of shape (n, embedding_dim)
+            The shared embedding.
+        """
+        return cast(
+            jnp.ndarray,
```
shaharbar1: Why is the casting needed? Doesn't it happend naturally?

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,314 @@
...
+                next_layer_input=x,
+                weights_biases=weights_biases,
+                activation_fn=self.jax_activation,
+                linear_fn=lambda a, w, b: jnp.dot(a, w) + b,
+                backend=jnp,
+                use_residual_connections=False,
+            ),
+        )
+
+    def update_params(self, weights: List[np.ndarray], biases: List[np.ndarray]) -> "MLPBackbone":
```
shaharbar1: the proper naming convention is "with_weights_and_biases". You should use the native pydantic model_copy with overided args

pybandits/model/bnn/network.py
```diff
@@ -1109,13 +1120,38 @@ def _build_svi_guide_init(self) -> tuple:
                 all_sigmas.append(emb.params["sigma"].ravel())
                 site_sigmas[emb_name] = emb.params["sigma"]
 
+        return values, site_sigmas, all_mus, all_sigmas
+
+    @staticmethod
+    def build_guide_init_fns(
```
shaharbar1: Why is it static? You could benefit from having self as parameter.

pybandits/base_model.py
```diff
@@ -163,8 +163,23 @@ def update(self, rewards: List[BinaryReward], **kwargs):
...
+        The canonical success/failure bookkeeping, factored out of ``update`` so callers that train
+        through a different path (e.g. the joint cMAB SVI engine, which bypasses per-model ``update``)
+        can keep the counters in sync without re-implementing the arithmetic.
+
+        Parameters
+        ----------
+        rewards : List[BinaryReward]
+            The binary reward for each sample.
+        """
+        successes = int(sum(rewards))
```
shaharbar1: Why do we need this line if rewards is List[BinaryReward]?

pybandits/base_model.py
```diff
@@ -163,8 +163,23 @@ def update(self, rewards: List[BinaryReward], **kwargs):
...
         """
         self._update(rewards=rewards, **kwargs)
-        self.n_successes += sum(rewards)
-        self.n_failures += len(rewards) - sum(rewards)
+        self._record_rewards(rewards)
+
+    def _record_rewards(self, rewards: List[BinaryReward]) -> None:
+        """Tally binary rewards into the success/failure counters.
+
+        The canonical success/failure bookkeeping, factored out of ``update`` so callers that train
```
shaharbar1, line 174: Good idea!

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,451 @@
...
+
+CmabHeadType = TypeVar(
+    "CmabHeadType",
+    bound=Union[BaseBayesianNeuralNetwork, BaseBayesianNeuralNetworkMO, BaseQuantitativeBayesianNeuralNetwork],
+)
+
+# Divider between an arm's scope prefix and its (otherwise identical) per-head site names in the
+# joint model. ``numpyro.handlers.scope`` prepends ``f"{prefix}{_SCOPE_DIVIDER}"`` to every site it
+# wraps, so the same value is used both to scope the model and to slice each arm's posterior back out.
+_SCOPE_DIVIDER = "/"
```
shaharbar1: Why not as class vars?

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,451 @@
...
+        quantities: Optional[List[Union[float, List[float], None]]] = None,
+        context: Optional[np.ndarray] = None,
+        **kwargs: Any,
+    ) -> None:
+        """Run one joint SVI pass over the batch (VI-only), then drop it (no replay buffer)."""
+        # VI-only is a property of the (consistency-validated) heads — reject MCMC up front, before
+        # touching any per-arm state, so a misconfigured meta-model fails with a clear message.
+        if self._rep_bnn(next(iter(self.actions.values()))).update_method != "VI":
+            raise ValueError("Contextual-bandit joint training is VI-only; set update_method='VI'.")
+        if context is None:
```
shaharbar1: Why is context Optional in the first place?

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,451 @@
...
+        actions: List[ActionId],
+        rewards: Union[List[BinaryReward], List[List[BinaryReward]]],
+        quantities: Optional[List[Union[float, List[float], None]]] = None,
+        context: Optional[np.ndarray] = None,
+        **kwargs: Any,
+    ) -> None:
+        """Run one joint SVI pass over the batch (VI-only), then drop it (no replay buffer)."""
+        # VI-only is a property of the (consistency-validated) heads — reject MCMC up front, before
+        # touching any per-arm state, so a misconfigured meta-model fails with a clear message.
+        if self._rep_bnn(next(iter(self.actions.values()))).update_method != "VI":
```
shaharbar1: This should be checked on init validations, not during update.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,451 @@
...
+                z = x_
+            for arm, (row_idx, arm_rewards) in arm_data_.items():
+                z_arm = z[row_idx]
+                for bnn, obj_index, is_quantitative in self._arm_units(self.actions[arm]):
+                    x_unit = jnp.concatenate([arm_q[arm], z_arm], axis=1) if is_quantitative else z_arm
+                    y_unit = arm_rewards if obj_index is None else arm_rewards[:, obj_index]
+                    # scope() prepends "{arm}/" (and "/objN" for MO) to every site this unit emits, so
+                    # the otherwise-identical per-head site names stay unique across arms.
+                    with numpyro.handlers.scope(prefix=self._unit_scope(arm, obj_index), divider=_SCOPE_DIVIDER):
+                        bnn.emit_submodel(x=x_unit, y=y_unit, kl_annealing_factor=cast(Any, kl_annealing_factor))
```
shaharbar1: great!

pybandits/meta_model.py
```diff
@@ -452,52 +479,6 @@ def reset(self) -> None:
             model.reset()
```
shaharbar1: I suggest making meta model a folder with 3 files: base.py, cmab_meta_model, smab_meta_model.

pybandits/meta_model.py
```diff
@@ -509,9 +490,3 @@ def check_models(self) -> "CmabPerActionMetaModel":
 PerActionMetaModelSmabDP = PerActionMetaModel[Union[BetaDP, ZoomingDP]]
 PerActionMetaModelSmabMO = PerActionMetaModel[BetaMO]
```
shaharbar1: So, essentially, did you kill the option for separate training of arms with cmab?

shaharbar1 (changes requested):

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,297 @@
...
+    # Uniform-init half-width numerator: limit = sqrt(numerator / fan), shared by He and Glorot/Xavier.
+    _init_limit_numerator: ClassVar[float] = 6.0
+    # Activations whose negative-half saturation calls for He init (variance ~ 2/fan_in).
+    _he_activations: ClassVar[Tuple[str, ...]] = ("relu", "gelu")
+
+    n_features: PositiveInt
+    embedding_dim: PositiveInt
+    hidden_dims: List[PositiveInt]
+    activation: ActivationFunctions = "relu"
+    random_seed: Optional[NonNegativeInt] = None
```
shaharbar1, line 65: If you have this, I would expect to have an internal _rng like in mab on _rng_key like in bnn

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,297 @@
...
+    weights: List[List[List[float]]]  # per layer: (input_dim, output_dim)
+    biases: List[List[float]]  # per layer: (output_dim,)
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
shaharbar1: For the sake of order, put this at the beginning before all attributes.

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,297 @@
...
+            Per-layer weight matrices.
+        biases : List[np.ndarray]
+            Per-layer bias vectors.
+
+        Returns
+        -------
+        Self
+            A new backbone carrying the supplied parameters.
+        """
+        return self.model_copy(
```
shaharbar1, line 292: Nice!

pybandits/model/bnn/backbone.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def reset(self) -> Self:
+        """Return a fresh backbone with the same architecture / seed (weights re-initialised).
+
+        Returns
+        -------
+        MLPBackbone
+            A cold-started backbone with this instance's configuration.
+        """
+        return self.cold_start(
```
shaharbar1, line 218: philosophical question: Are we cool with the fact that each reset will create different initial weights?
I think we do. WDYT?

pybandits/model/bnn/network.py
```diff
@@ -725,73 +698,101 @@ def _create_update_model(self) -> Callable:
...
+            self.emit_submodel(x=x, y=y, kl_annealing_factor=kl_annealing_factor, batch_size=batch_size)
+
+        return model
+
+    def emit_submodel(
+        self,
+        x: jax.Array,
+        y: jax.Array,
+        kl_annealing_factor: Union[PositiveFloat01, jax.Array] = 1.0,
+        batch_size: Optional[int] = None,
```
shaharbar1: Optional[PositiveInt]

pybandits/cmab.py
```diff
@@ -301,6 +304,11 @@ class CmabBernoulli(BaseCmabBernoulli):
     strategy: ClassicBandit
     _predict_with_proba: bool = False
 
+    # ``cold_start`` is inherited from ``BaseMab``: a shared backbone is requested by passing
```
shaharbar1: It's a good comment, but that's not the best place for it

pybandits/meta_model.py
shaharbar1: Remember this? It is highly relevant

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,442 @@
...
+        through the generic ``BaseMab.cold_start`` factory, so no bespoke cmab ``cold_start`` is needed.
+
+        No ``Union`` lives on the manager field for this class, so a custom ``__init__`` is safe.
+        """
+        # Mutate the caller's kwargs in place (don't copy): the bandit factory relies on the
+        # consumed keys disappearing from this dict to detect leftover unknown arguments.
+        kwargs = kwargs or {}
+        if backbone is None:
+            backbone = type(self)._build_backbone_from_kwargs(kwargs, random_seed)
+        actions_dict = type(self)._preprocess_actions(
```
shaharbar1: Doesn't this already happen on super().__init__?

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,442 @@
...
+        ``embedding_dim``/``backbone_activation``) — the latter is how ``cold_start`` reaches this class
+        through the generic ``BaseMab.cold_start`` factory, so no bespoke cmab ``cold_start`` is needed.
+
+        No ``Union`` lives on the manager field for this class, so a custom ``__init__`` is safe.
+        """
+        # Mutate the caller's kwargs in place (don't copy): the bandit factory relies on the
+        # consumed keys disappearing from this dict to detect leftover unknown arguments.
+        kwargs = kwargs or {}
+        if backbone is None:
+            backbone = type(self)._build_backbone_from_kwargs(kwargs, random_seed)
```
shaharbar1: why do we need type(self)? why not self?

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,442 @@
...
+        actions_dict = type(self)._preprocess_actions(
+            actions=actions,
+            action_ids=action_ids,
+            quantitative_action_ids=quantitative_action_ids,
+            kwargs=kwargs,
+        )
+        super().__init__(actions=actions_dict, backbone=backbone, random_seed=random_seed)  # type: ignore[call-arg]
+
+    @classmethod
+    def _build_backbone_from_kwargs(cls, kwargs: Dict[str, Any], random_seed: Optional[int]) -> Optional[MLPBackbone]:
```
shaharbar1: random seed should be non negative int, not int

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,442 @@
...
+        a serialized dict) or requested through ``kwargs`` via ``backbone_hidden_dims`` (plus optional
+        ``embedding_dim``/``backbone_activation``) — the latter is how ``cold_start`` reaches this class
+        through the generic ``BaseMab.cold_start`` factory, so no bespoke cmab ``cold_start`` is needed.
+
+        No ``Union`` lives on the manager field for this class, so a custom ``__init__`` is safe.
+        """
+        # Mutate the caller's kwargs in place (don't copy): the bandit factory relies on the
+        # consumed keys disappearing from this dict to detect leftover unknown arguments.
+        kwargs = kwargs or {}
+        if backbone is None:
```
shaharbar1: backbone = backbone or ....

pybandits/cmab_meta_model.py
```diff
@@ -0,0 +1,442 @@
...
+    random_seed: Optional[NonNegativeInt] = None
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    # Default backbone output width when ``embedding_dim`` is omitted: n_features // divisor (min 1),
+    # mirroring the BNN's categorical-embedding sizing (``_embedding_dim_divisor``).
+    _embedding_dim_divisor: ClassVar[int] = 4
+    # Divider between an arm's scope prefix and its per-head site names in the joint NumPyro model.
+    # ``numpyro.handlers.scope`` prepends ``f"{prefix}{_SCOPE_DIVIDER}"`` to every site it wraps.
+    _SCOPE_DIVIDER: ClassVar[str] = "/"
```
shaharbar1: no need for capital letters

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,442 @@
...
+    ) -> None:
+        """Run one joint SVI pass over the batch (VI-only), then drop it (no replay buffer)."""
+        # Trim memory rows (shared with PerActionMetaModel); the manager already cast context to float.
+        context, quantities = self._align_to_actions(actions, context, quantities)
+        expected_cols = (
+            self.backbone.n_features
+            if self.backbone is not None
+            else self._context_dim(next(iter(self.actions.values())))
+        )
+        if context.shape[1] != expected_cols:
```
shaharbar1: isn't this already done somewhere before? I guess in the actionsmanager

pybandits/meta_model.py
```diff
@@ -186,6 +178,57 @@ def update(
...
+    @staticmethod
+    def _check_action_count(actions: Dict[ActionId, BaseModel]) -> None:
+        """Shared action-count guard: at least one action; a single action is deterministic."""
+        if len(actions) == 0:
+            raise AttributeError("At least one action should be defined.")
+        if len(actions) == 1:
+            warnings.warn("Only a single action was supplied. This MAB will be deterministic.")
+
+    @staticmethod
+    def _align_to_actions(
```
shaharbar1: You don't need this and it can silence data issues, e.g. with the trimming [:-n]. The way I recall this is done on the actions manager

shaharbar1 (changes requested):

pybandits/meta_model.py
AnastasiiaKabeshova: agree, removed the trim

pybandits/meta_model/cmab_meta_model.py
AnastasiiaKabeshova: No , the manager's _check_context_matrix only casts to float; it never validates the feature-column count

pybandits/model/bnn/backbone.py
AnastasiiaKabeshova: With a random_seed, reset() derives identical weights (deterministic). Only random_seed=None varies per reset ‚ matching BNN/MAB. So yes, I think we're fine with it.

pybandits/model/bnn/backbone.py
AnastasiiaKabeshova: The backbone is deterministic ‚ trained weights are stored point estimates, optimised as numpyro.param. random_seed is used only for reproducible init at cold_start/reset. A persistent RNG would be dead state that never advances. I suggest keep random_seed, no RNG.

pybandits/meta_model/smab_meta_model.py
```diff
@@ -0,0 +1,140 @@
...
+import numpy as np
+from pydantic import field_validator
+
+from pybandits.base import ActionId, BinaryReward
+from pybandits.meta_model.base import ActionModelType, BaseMetaModel, SampleProbaResult
+from pybandits.model import Beta, BetaCC, BetaDP, BetaMO, BetaMOCC
+from pybandits.quantitative_model import Zooming, ZoomingCC, ZoomingDP
+
+
+class PerActionMetaModel(BaseMetaModel, Generic[ActionModelType]):
```
shaharbar1: Either change class name to SmabMetaModel, or file name to per_action_meta_model.py. I think first is better.

pybandits/meta_model/base.py
```diff
@@ -123,6 +96,29 @@ class BaseMetaModel(PyBanditsBaseModel, ABC):
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
+    def __init__(
+        self,
+        actions: Optional[Dict[ActionId, BaseModel]] = None,
+        action_ids: Optional[Set[ActionId]] = None,
+        quantitative_action_ids: Optional[Set[ActionId]] = None,
+        kwargs: Optional[Dict[str, Any]] = None,
```
shaharbar1, line 110: I think you can have extra and kwargs under **kwargs. _preprocess_actions should grab what it needs from kwargs and pop it. Remainders would go to init.

pybandits/meta_model/base.py
```diff
@@ -186,6 +182,43 @@ def update(
...
+    @staticmethod
+    def _check_action_count(actions: Dict[ActionId, BaseModel]) -> None:
+        """Shared action-count guard: at least one action; a single action is deterministic."""
+        if len(actions) == 0:
+            raise AttributeError("At least one action should be defined.")
+        if len(actions) == 1:
+            warnings.warn("Only a single action was supplied. This MAB will be deterministic.")
+
+    @staticmethod
+    def _check_lengths_match(
```
shaharbar1: already happens under _validate_update_params in actions manager...

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+        if self.backbone is not None and first_dim != self.backbone.embedding_dim:
+            raise AttributeError(
+                f"Head context dim ({first_dim}) must equal backbone.embedding_dim ({self.backbone.embedding_dim})."
+            )
+        if self._rep_bnn(heads[0]).update_method != "VI":
+            raise ValueError("Contextual-bandit joint training is VI-only; set update_method='VI'.")
+        return self
+
+    # --------------------------------------------------------------------- head-shape helpers
+    # A "head" is one of: a single BNN (SO/CC/DP), an MO head (a list of per-objective BNNs), or a
```
shaharbar1: I think this comment should be in the class docstring...

pybandits/meta_model/cmab_meta_model.py
shaharbar1: Correct!

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+        self,
+        actions: List[ActionId],
+        rewards: Union[List[BinaryReward], List[List[BinaryReward]]],
+        context: np.ndarray,
+        quantities: Optional[List[Union[float, List[float], None]]] = None,
+        **kwargs: Any,
+    ) -> None:
+        """Run one joint SVI pass over the batch (VI-only), then drop it (no replay buffer)."""
+        # Rows are aligned by the manager (context already cast to float); guard against a raw mismatch.
+        self._check_lengths_match(actions, context, quantities)
```
shaharbar1: Already done in  _validate_update_params in actionsmanager

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+        expected_cols = (
```
shaharbar1: make this a property for convinience?

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+
+    # --------------------------------------------------------------------- joint SVI engine
+    def _joint_svi_update(
+        self,
+        context: np.ndarray,
+        arm_to_rows: Dict[ActionId, List[int]],
+        rewards_arr: np.ndarray,
+        quantities: Optional[List[Union[float, List[float], None]]],
+    ) -> None:
+        """Train the (optional) backbone point estimates + all per-arm head posteriors in one SVI pass."""
```
shaharbar1: Elaborate this docstring, this is the core of the change!

Add explaination on how the training happens jointly. Add parameters sections

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+        batch_arms = list(arm_to_rows)
+        rep = self._rep_bnn(self.actions[batch_arms[0]])
+        if rep._update_kwargs["method"] != "advi":
```
shaharbar1: This should be on class validator method, not here.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+                for bnn, obj_index, is_quantitative in self._arm_units(self.actions[arm]):
+                    x_unit = jnp.concatenate([arm_q[arm], z_arm], axis=1) if is_quantitative else z_arm
+                    y_unit = arm_rewards if obj_index is None else arm_rewards[:, obj_index]
+                    # scope() prepends "{arm}/" (and "/objN" for MO) to every site this unit emits, so
+                    # the otherwise-identical per-head site names stay unique across arms.
+                    with numpyro.handlers.scope(prefix=self._unit_scope(arm, obj_index), divider=self._scope_divider):
+                        bnn.emit_submodel(x=x_unit, y=y_unit, kl_annealing_factor=cast(Any, kl_annealing_factor))
+
+        values, site_sigmas, all_mus, all_sigmas = {}, {}, [], []
+        for arm in batch_arms:
```
shaharbar1, line 713: Put this in dedicated function for readability. The function of joint update is long enough to follow.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+                site_sigmas.update({f"{scope}{name}": val for name, val in s.items()})
+                all_mus += mu
+                all_sigmas += sg
+        init_loc_fn, init_scale_fn = rep.build_guide_init_fns(values, site_sigmas, all_mus, all_sigmas)
+        guide = ParameterizedScaleAutoNormal(_model, init_loc_fn=init_loc_fn, init_scale_fn=init_scale_fn)
+
+        # Full-batch joint pass; epoch/batch_size/num_steps settings are read from rep._update_kwargs
+        # via the shared _compute_epoch_steps helper — same precedence rules as the standalone BNN path.
+        n_samples = len(context)
+        epoch_factor_arrays = rep._build_kl_annealing_factors(rep._compute_epoch_steps(n_samples))
```
shaharbar1: make _build_kl_annealing_factors unprotected

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+        svi, params, _history, self._rng_key = run_svi(
+            model=_model,
+            guide=guide,
+            optimizer=rep._obj_optimizer,
+            loss=loss,
+            rng_key=self._rng_key,
+            model_args=(x, arm_data),
+            epoch_factor_arrays=epoch_factor_arrays,
+            restore_best=rep._update_kwargs.get("restore_best_svi_state", True),
+            desc="cmab joint SVI",
```
shaharbar1: Why do we need this desc field?
Anyhow, put this in a class var if you need it, don't hard code the string

pybandits/meta_model/cmab_meta_model.py
shaharbar1: use update_kwargs, not _update_kwargs

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+    def _joint_svi_update(
+        self,
+        context: np.ndarray,
+        arm_to_rows: Dict[ActionId, List[int]],
+        rewards_arr: np.ndarray,
+        quantities: Optional[List[Union[float, List[float], None]]],
+    ) -> None:
+        """Train the (optional) backbone point estimates + all per-arm head posteriors in one SVI pass."""
+        batch_arms = list(arm_to_rows)
+        rep = self._rep_bnn(self.actions[batch_arms[0]])
```
shaharbar1: I think rep should be a private attribute of the class. give him meaningfull name like representative_bnn

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+        self.backbone = self.backbone.with_weights_and_biases(new_w, new_b)
+
+    def _increment_counters(self, arm_to_rows: Dict[ActionId, List[int]], rewards_arr: np.ndarray) -> None:
+        """Keep each head's success/failure counters in sync (the joint engine bypasses per-head ``update``).
+
+        These counters feed the manager's adaptive-window stats. The joint SVI pass trains every arm at
+        once instead of calling each head's ``update``, so we replay the rewards through the canonical
+        :meth:`BaseModelSO._record_rewards` bookkeeping: the success/failure owner is the per-objective
+        sub-model for MO heads and the head itself (a ``BaseModelSO``) for single-objective/quantitative.
+        """
```
shaharbar1, line 821: In general, your docstrings miss parameters section and returns section (only when relevant)

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,447 @@
...
+        for arm, rows in arm_to_rows.items():
+            head = self.actions[arm]
+            if isinstance(head, BaseBayesianNeuralNetworkMO):
+                arm_rewards = rewards_arr[rows]
+                for i, sub in enumerate(head.models):
+                    sub._record_rewards(list(arm_rewards[:, i]))
+            else:  # BNN or quantitative head — both BaseModelSO that own their counters
+                head._record_rewards(list(np.asarray(rewards_arr[rows]).reshape(-1)))
```
shaharbar1: un-protect the method _record_rewards.
I still don't get why you need this here and rely on the original calls for _record_rewards

pybandits/model/bnn/backbone.py
shaharbar1: Cool

pybandits/meta_model/cmab_meta_model.py
shaharbar1: OK, I get why you need it :-) just do it record_rewards instead of _record_rewards

pybandits/cmab.py
```diff
@@ -78,6 +78,9 @@ class BaseCmabBernoulli(BaseMab, ABC):
     @property
     def input_dim(self) -> int:
         """Returns the input feature dimension (number of context features)."""
+        meta_model = self.actions_manager.meta_model
```
shaharbar1: put an input_dim property with this calc inside the cmab meta model. Then just pull input_dim from there. the met model should manage this calculation

tests/test_cmab.py
```diff
@@ -1256,3 +1250,107 @@ def test_random_seed_propagates_to_bnn(
     assert actions1 == actions2
     assert probs1 == probs2
     assert ws1 == ws2
+
+
+# --- shared-backbone (encoder + per-arm heads) cold_start test constants ---
```
shaharbar1: Put the new tests into a class of tests.
Make the constants class vars and uncapitalize them.
Use hypothesis.given instead of constants when applicable to earn better maintainability of tests.

tests/test_cmab_meta_model.py
```diff
@@ -0,0 +1,242 @@
...
+ABSENT_ARM: ActionId = "c"
+N_OBJECTIVES = 2
+QUANTITY_DIM = 1
+NUM_STEPS = 15
+BACKBONE_SEED = 7
+META_SEED = 13
+SHAPE_TEST_SEED = 0
+
+
+def _uk() -> dict:
```
shaharbar1: Use a better name

tests/test_cmab_meta_model.py
```diff
@@ -0,0 +1,242 @@
...
+import numpy as np
+import pytest
+from hypothesis import given, settings
+from hypothesis import strategies as st
+
+from pybandits.base import ActionId
+from pybandits.meta_model import CmabMetaModelMO, CmabMetaModelSO
+from pybandits.model.bnn.backbone import MLPBackbone
+
+N_FEATURES = 6
```
shaharbar1: SAme comment as for the former tests

tests/test_cmab_meta_model.py
```diff
@@ -0,0 +1,242 @@
...
+    arms = sorted(ACTION_IDS)
+    actions = [arms[i % len(arms)] for i in range(n_rows)]
+    rewards = rng.integers(0, 2, size=n_rows).tolist()
+    return actions, rewards, _context(rng, n_rows)
+
+
+# --------------------------------------------------------------------------- MLPBackbone
+
+
+@settings(deadline=None, max_examples=20)
```
shaharbar1: You don't need settings here...

tests/test_cmab_meta_model.py
```diff
@@ -0,0 +1,242 @@
...
+BACKBONE_SEED = 7
+META_SEED = 13
+SHAPE_TEST_SEED = 0
+
+
+def _uk() -> dict:
+    """Fresh head cold-start kwargs (the cold-start machinery pops consumed keys)."""
+    return {"n_features": N_FEATURES, "hidden_dim_list": HIDDEN, "update_kwargs": {"num_steps": NUM_STEPS}}
+
+
```
shaharbar1, line 47: Use fixtures when applicable.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pybandits"
-version = "7.4.1"
+version = "7.4.0"
```
shaharbar1: 8.0.0?

shaharbar1 (changes requested):

pybandits/meta_model/base.py
```diff
@@ -196,7 +215,7 @@ def _preprocess_actions(
     ) -> Dict[ActionId, BaseModel]:
         """Return the canonical per-action dict the meta-model should store.
 
-        Default behaviour (used by ``PerActionMetaModel`` and any subclass whose state is
+        Default behaviour (used by ``SmabMetaModel`` and any subclass whose state is
```
shaharbar1: If it is only used by the child class -> make it abstract here, and define it only on the child classes.

pybandits/meta_model/base.py
```diff
@@ -123,6 +96,44 @@ class BaseMetaModel(PyBanditsBaseModel, ABC):
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
+    def __init__(
+        self,
+        actions: Optional[Dict[ActionId, BaseModel]] = None,
+        action_ids: Optional[Set[ActionId]] = None,
+        quantitative_action_ids: Optional[Set[ActionId]] = None,
+        kwargs: Optional[Dict[str, Any]] = None,
+        **extra: Any,
```
shaharbar1: Remove extra.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,583 @@
...
+    QuantitativeBayesianNeuralNetworkDP,
+)
+
+CmabHeadType = TypeVar(
+    "CmabHeadType",
+    bound=Union[BaseBayesianNeuralNetwork, BaseBayesianNeuralNetworkMO, BaseQuantitativeBayesianNeuralNetwork],
+)
+
+
+class CmabMetaModel(BaseMetaModel, Generic[CmabHeadType]):
```
shaharbar1, line 77: The code looks good!

tests/test_transfer.py
```diff
@@ -611,6 +610,8 @@ def test_transfer_extendable_key_change_warns(
     ) -> None:
         """Test that changing an extendable key emits a warning but succeeds (parametrized over transfer_extendable_keys)."""
         caplog.set_level(logging.WARNING)
+        if key == "update_method" and val2 == "MCMC":
```
shaharbar1: Did you block MCMC?

pybandits/actions_manager.py
```diff
@@ -149,7 +149,7 @@ def _get_meta_model_cls(cls) -> Type[BaseMetaModel]:
             if type_args and not isinstance(type_args[0], TypeVar):
                 return annotation  # type: ignore[return-value]
             return get_origin(annotation) or annotation  # type: ignore[return-value]
-        return PerActionMetaModel
+        return SmabMetaModel
```
shaharbar1: I don't like this. raise an error instead of returning something not consistent.

tests/test_cmab.py
```diff
@@ -1256,3 +1250,120 @@ def test_random_seed_propagates_to_bnn(
     assert actions1 == actions2
     assert probs1 == probs2
     assert ws1 == ws2
+
+
+class TestCmabSharedBackbone:
```
shaharbar1: Why a new class of tests when you can use the existing tests and just add backbone parameter (make it None or some value like [2])

tests/test_transfer.py
shaharbar1: You can remove it from network.py...

shaharbar1 (changes requested):

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,781 @@
...
+
+        Returns
+        -------
+        Callable
+            The NumPyro model ``(x, action_index, y, kl_annealing_factor) -> None``.
+        """
+        activation_fn = representative_bnn._jax_activation_fn
+        use_residual = representative_bnn.use_residual_connections
+
+        def einsum_linear(t: jax.Array, w: jax.Array, b: jax.Array) -> jax.Array:
```
shaharbar1: You have a lot of functions inside functions, I don't like it, it is harder to read and follow and slower in runtime. avoid it when possible.

pybandits/meta_model/smab_meta_model.py
```diff
@@ -0,0 +1,138 @@
...
+    Domain invariants (e.g. same number of objectives for smab; consistent
+    input size / update method for cmab) are validated on the respective manager
+    subclasses, not here.
+    """
+
+    actions: Dict[ActionId, ActionModelType]  # type: ignore[valid-type]
+
+    @field_validator("actions", mode="after")
+    @classmethod
+    def at_least_one_action_is_defined(cls, v: Dict[ActionId, ActionModelType]) -> Dict[ActionId, ActionModelType]:
```
shaharbar1: Isn't this already happening on the father class? If not, then it should

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,781 @@
...
+
+    actions: Dict[ActionId, CmabHeadType]  # type: ignore[assignment]
+    backbone: Optional[MLPBackbone] = None
+    random_seed: Optional[NonNegativeInt] = None
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    # Default backbone output width when ``embedding_dim`` is omitted: n_features // divisor (min 1),
+    # mirroring the BNN's categorical-embedding sizing (``_embedding_dim_divisor``).
+    _embedding_dim_divisor: ClassVar[int] = 4
```
shaharbar1: This is defined here and on bnn. Define it only here as unprotected class var and have the bnn use it

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,781 @@
...
+            embedding_dim=emb,
+            activation=activation or "relu",
+            random_seed=kwargs.get("random_seed", random_seed),
+        )
+        kwargs["n_features"] = emb  # per-arm heads live on the embedding, not the raw context
+        return backbone
+
+    def model_post_init(self, __context: Any) -> None:
+        """Seed the JAX PRNG key used to drive the joint SVI pass."""
+        self._rng_key = jax.random.PRNGKey(self.random_seed if self.random_seed is not None else 0)
```
shaharbar1: In bnn, do we also replace with 0 when it is none?

shaharbar1 (approved):

pybandits/meta_model/base.py
```diff
@@ -186,28 +200,65 @@ def update(
...
+        context : Optional[np.ndarray]
+            Per-sample context rows, row-aligned with ``actions``; ``None`` for smab.
         """
-        return cls._instantiate_actions(
-            actions=actions,
-            action_ids=action_ids,
-            quantitative_action_ids=quantitative_action_ids,
-            kwargs=kwargs,
-        )
+        rewards_dict: Dict[ActionId, List[Any]] = defaultdict(list)
```
shaharbar1, line 238: the Dict typehint is wrong when you define it as defaultdict. I think you should remove the typehint here.

pybandits/meta_model/base.py
```diff
@@ -186,28 +200,65 @@ def update(
...
+        context_dict: Dict[ActionId, List[Any]] = defaultdict(list)
+
+        if quantities is None:
+            for i, (a, r) in enumerate(zip(actions, rewards)):
+                rewards_dict[a].append(r)
+                if context is not None:
```
shaharbar1: I don't like the optional pattern. This behvior is cmab specific, so it shouldn't be on the base class. I would suggest to make this function abstract with **kwargs instead of context on the base. Then iplement specifically for each of the sons.

pybandits/meta_model/base.py
```diff
@@ -317,7 +368,7 @@ def _action_model_classes(cls) -> Tuple[Type[BaseModel], ...]:
         if not type_args or len(type_args) < 2:
             raise TypeError(
                 f"{cls.__name__}.actions has no concrete value type annotation; "
-                "parameterise the meta-model (e.g. PerActionMetaModel[Beta]) before cold-starting."
+                "parameterise the meta-model (e.g. SmabMetaModel[Beta]) before cold-starting."
```
shaharbar1: remove the e.g., Don't refer to the child on base

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,830 @@
...
+            if set_params:
+                raise TypeError(f"{set_params} only apply with a backbone; pass backbone_hidden_dims=[...].")
+            return None
+        n_features = kwargs.get("n_features")
+        if n_features is None:
+            raise ValueError("n_features must be provided to build the shared backbone.")
+        emb = (
+            embedding_dim
+            if embedding_dim is not None
+            else max(1, n_features // BaseBayesianNeuralNetwork.embedding_dim_divisor)
```
shaharbar1, line 177: Maybe embedding_dim_divisor should be on DNNMixin? not sure

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,830 @@
...
+        # Rows are aligned by the manager (_validate_update_params), which also cast context to float.
+        if context.shape[1] != self.input_dim:
+            raise AttributeError(f"context has {context.shape[1]} feature columns; expected {self.input_dim}.")
+        rewards_arr: np.ndarray = np.asarray(rewards)
+        # Multi-objective heads need 2-D rewards with one column per objective (the per-model update
+        # used to validate this; both the joint engine and the independent per-arm dispatch bypass
+        # that per-model check, so it is enforced here regardless of backbone).
+        for action_id in set(actions):
+            head = self.actions[action_id]
+            if isinstance(head, BaseBayesianNeuralNetworkMO):
```
shaharbar1, line 345: Isn't this already checked on actionsmanager?

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,830 @@
...
+            z_arm = z[row_idx]
+            for bnn, obj_index, is_quantitative in self._arm_units(self.actions[arm]):
+                x_unit = jnp.concatenate([arm_q[arm], z_arm], axis=1) if is_quantitative else z_arm
+                y_unit = arm_rewards if obj_index is None else arm_rewards[:, obj_index]
+                # scope() prepends "{arm}/" (and "/objN" for MO) to every site this unit emits, so
+                # the otherwise-identical per-head site names stay unique across arms.
+                with numpyro.handlers.scope(prefix=self._unit_scope(arm, obj_index), divider=self._scope_divider):
+                    bnn.emit_submodel(x=x_unit, y=y_unit, kl_annealing_factor=cast(Any, kl_annealing_factor))
+
+    def _minibatch_applicable(self, batch_size: Optional[int], n_samples: int) -> bool:
```
shaharbar1: Feels to me like this should be a property without all the arguments. checking if batch_size is None or batch_size >= n_samples:, should be done outside, it doesn't have to on whether the model is minibatch applicable.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,830 @@
...
+                weights_biases=per_sample_wb,
+                activation_fn=representative_bnn._jax_activation_fn,
+                linear_fn=self._einsum_linear,
+                backend=jnp,
+                use_residual_connections=representative_bnn.use_residual_connections,
+            ).squeeze(-1)
+            numpyro.sample("out", NumpyroBernoulli(logits=logit), obs=y_[idx])
+
+    @staticmethod
+    def _einsum_linear(t: jax.Array, w: jax.Array, b: jax.Array) -> jax.Array:
```
shaharbar1: Isn't this relevant for DNNMixin or _svi.py? I mean can't it be used by BNN as well?

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -0,0 +1,830 @@
...
+        return functools.partial(
+            self._minibatch_model,
+            batch_arms=batch_arms,
+            representative_bnn=representative_bnn,
+            batch_size=batch_size,
+            n_samples=n_samples,
+            declare_backbone=declare_backbone,
+        )
+
+    def _minibatch_model(
```
shaharbar1: How is this so different from _full_batch_model? Can't emit_model be used here as well?

shaharbar1: I would have expected _full_batch_model and _minibatch_model to be more similar through emit_model or an internal function used by emit model. I'm again afraid of code repetitions.
Ultimately, my hunch is that the two functions can be united via the plate mechanism like in the bnn. The plating for the heads is joint for the to functions, only the data plating is nor the same. It should be handled like on the bnn code via nullcontext

pybandits/model/bnn/network.py
```diff
@@ -1063,14 +1094,36 @@ def _build_svi_guide_init(self) -> tuple:
                 all_sigmas.append(emb.params["sigma"].ravel())
                 site_sigmas[emb_name] = emb.params["sigma"]
 
+        return values, site_sigmas, all_mus, all_sigmas
+
+    def build_guide_init_fns(self, values: dict, site_sigmas: dict, all_mus: list, all_sigmas: list) -> tuple:
```
shaharbar1, line 1098: very nice OOP

pybandits/actions_manager.py
```diff
@@ -143,13 +143,16 @@ def actions(self) -> Dict[ActionId, BaseModel]:
...
         """Return the concrete meta-model class from the manager's ``meta_model`` field annotation."""
         meta_model_field = cls.model_fields.get("meta_model")
-        if meta_model_field is not None and meta_model_field.annotation is not None:
-            annotation = meta_model_field.annotation
-            type_args = get_args(annotation)
-            if type_args and not isinstance(type_args[0], TypeVar):
-                return annotation  # type: ignore[return-value]
-            return get_origin(annotation) or annotation  # type: ignore[return-value]
-        return PerActionMetaModel
+        if meta_model_field is None or meta_model_field.annotation is None:
```
shaharbar1: Why would this happen? isn't this an overkill? I would expect pydantic to catch it on its own...

shaharbar1: If you allowed meta model as optional on the actions_manager, then just don't... and save this validation

shaharbar1 (approved):

PR #149, opened 2026-06-17
Allow quantitative regions exclusion

shaharbar1:
### Changes:

* Add ForbiddenRegion and ForbiddenActions type aliases to pybandits/base.py — signed-margin region callables (>0 forbidden) and generalised dict form supporting whole-arm and per-arm region blocking
* Add _normalize_forbidden_actions() to BaseMab — canonicalises Set/dict forbidden forms into valid_actions and per-arm feasibility constraints; validates quantitative-only regions and default-action placement
* Add _to_feasibility_constraint() static method to BaseMab — negates signed-margin region into optimizer-compatible ">= 0 feasible" form
* Add _sample_allowed_quantity() to BaseMab — rejection-sampler for the explore branch; drops an arm from explore candidates if its entire quantity space is forbidden within max_tries
* Update _select_epsilon_greedy_action() in BaseMab — accepts forbidden_regions; region-aware explore branch rejection-samples allowed quantities and falls back to strategy when all candidates are exhausted
* Update _get_action_probabilities() signature in BaseMab — takes pre-computed valid_actions instead of raw forbidden_actions (normalization lifted to callers)
* Update predict() in BaseMab, BaseSmabBernoulli, BaseCmabBernoulli — accepts ForbiddenActions; delegates normalization to _normalize_forbidden_actions before sampling
* Update select_action() and refine_p() in SingleObjectiveStrategy — propagates forbidden_regions into constrained quantity optimization
* Update select_action(), _get_feasible_solutions(), _get_pareto_front(), _get_approximate_pareto_front() in MultiObjectiveStrategy — filters Pareto candidate points landing inside forbidden regions before Pareto computation
* Move Segment class from pybandits/quantitative_model/zooming.py to pybandits/quantitative_model/segment.py — re-exported via __init__.py to preserve public API
* Add tests in tests/test_smab.py — covers all normalization forms, discrete-arm region rejection, default-action-in-region validation, feasibility constraint sign, rejection sampling, and explore/exploit region compliance
* Add tests in tests/test_strategy.py — covers refine_p steering quantity below region boundary and dropping fully-forbidden quantitative arms

## Summary by CodeRabbit

* **New Features**
  * Added region-based forbidden action constraints for quantitative actions, enabling partial blocking via predicate callables.
  * Introduced a `Segment` model to represent hyper-rectangles and generate region “outside” forbidden margins.

* **API Updates**
  * Expanded `forbidden_actions` handling across bandit `predict()` for both whole-arm and region-restricted quantitative constraints.
  * Updated strategies to accept per-action `forbidden_regions` plus optional RNG for deterministic quantitative refinement.
  * Added `seed` support to `maximize_by_quantity`.

* **Tests**
  * Added/extended coverage for normalization, validation, feasibility, sampling, and enforcement across explore/exploit paths.

* **Documentation**
  * Refreshed the CMAB quantitative notebook to demonstrate forbidden-region usage.

* **Chores**
  * Version bump to `7.5.0`.


pybandits/mab.py
```diff
@@ -357,11 +432,47 @@ def get_state(self) -> (str, str):
         state = self.model_dump_json()
         return model_name, state
 
+    def _sample_allowed_quantity(
```
ronshiff1, line 435: For a "simple" case - we could do something much more elegnat...
not sure it helps...
its quite easy to have a uniform sample over it.
this is how it can be done:
(without iterations)
Image

ronshiff1 (changes requested):

pybandits/strategy/multi_objective.py
```diff
@@ -230,14 +241,23 @@ def _get_approximate_pareto_front(
                 pareto_input_points = self._find_pareto_front_normal_constraint(
                     prob_or_func, actions[action_id].dimension, n_objectives, n_divisions, actions[action_id]
                 )
+                # The Normal-Constraint subproblems do not see forbidden regions, so drop any generated point that
+                # lands in this arm's forbidden region (a point is allowed only where every constraint is >= 0).
+                arm_constraints = forbidden_regions.get(action_id) if forbidden_regions else None
```
ronshiff1: this is problematic - can't we constraint the pareto front to be with only valid values? 
(send the constraints to the optimization)
we're risking that the pareto front is out of the allowed regions

pybandits/strategy/multi_objective.py
```diff
@@ -538,6 +558,8 @@ def _get_pareto_front(
         self,
         p: Dict[ActionId, Union[List[float], List[Callable[[np.ndarray], float]]]],
         actions: Dict[ActionId, BaseModel],
+        forbidden_regions: Optional[Dict[ActionId, List[Callable[[np.ndarray], float]]]] = None,
+        rng: Optional[Any] = None,
```
ronshiff1, line 577: rng is not used in the optimization below...
(also not in docstring)

pybandits/strategy/single_objective.py
ronshiff1: As far as I understand (and also claude...) we don't take into account the forbidden regions - which is  a problem 
I also don't understand what was the purpose of the row
constraint = constraint_list[0] if constraint_list else None
(above it)

pybandits/base.py
```diff
@@ -60,6 +62,20 @@
 QuantitativeMOProbability = Callable[[np.ndarray], MOProbability]
 QuantitativeMOProbabilityWeight = Tuple[Callable[[np.ndarray], MOProbability], Callable[[np.ndarray], float]]
 
+# A forbidden region restricts a quantitative action's quantity space [0, 1]^d.
+# Signed-margin convention: forbidden(x) > 0 => x is forbidden, <= 0 => x is allowed. A float margin (not a bare
+# bool) is required so the optimizer has directional information and degrades gracefully near the boundary.
+ForbiddenRegion = Callable[[np.ndarray], float]
```
ronshiff1, line 68: I think that in most cases - we're going to get an "allowed" region - 
meaning [0.2, 0.6] for example. 
what do you think about added an option to get a that this as input 
(and then when we normalize it - convert it to the functions as we expect)
that will simplify integration with optx

pybandits/mab.py
shaharbar1: Indeed more elegant.
However, I don't think that with the current optimization method it will make difference. In addition, it would require major refactor just for making correspondence with this proposal...

pybandits/base.py
shaharbar1: I agree, that's why I added forbidden_region_outside on Segment -> to convert allowed area to a forbidden region.

ronshiff1: is it also valid in the QBNN?

ronshiff1 (approved):

PR #151, opened 2026-06-21
Type BNN update_kwargs as validated Pydantic models

shaharbar1:
### Changes:

* Add VIUpdateKwargs and MCMCUpdateKwargs to pybandits/model/bnn/config.py — typed replacements for the untyped update_kwargs dict
  - constrained known fields (num_steps, method, num_particles, kl_annealing_fraction, num_warmup, num_samples, ...); optimizer_kwargs/lr_scheduler_kwargs/early_stopping_kwargs/nuts left as open passthrough dicts
  - MCMCUpdateKwargs inherits extra="forbid", which is what rejects VI-only kwargs under MCMC
  - VIUpdateKwargs warns (model_validator mode="before") when both epochs and num_steps are supplied, reading raw input so model_dump() round-trips do not warn spuriously
* Update BaseBayesianNeuralNetwork.update_kwargs to Optional[Union[VIUpdateKwargs, MCMCUpdateKwargs]] in pybandits/model/bnn/network.py
  - add _coerce_update_kwargs (model_validator mode="before") that selects the schema by update_method and copies the input dict to avoid leaking an injected key into sibling models during Union resolution
  - flatten the validated model via model_dump() into _update_kwargs so the internal training code stays unchanged
  - switch _get_obj_optimizer and _get_early_stopping_callback to attribute access
* Remove _arrange_update_kwargs and the _default_vi_kwargs/_default_mcmc_kwargs/_vi_update_params class-vars from network.py — superseded by the typed models
  - drop now-unused numbers and warnings imports
* Update kl_annealing_fraction tests in tests/test_model.py — drop the string/bool rejection cases so the field coerces like sibling numeric fields, keep the (0, 1] range checks
  - fix stale _arrange_update_kwargs reference in the MCMC-rejects test docstring

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bayesian Neural Network update parameter validation with stricter type checking for VI and MCMC configurations, providing clearer error messages for misconfigured parameters.

* **Tests**
  * Updated validation tests for KL-annealing fraction and parameter configuration constraints.

### DIFF
--- a/pybandits/model/bnn/config.py
+++ b/pybandits/model/bnn/config.py
@@ -326,8 +326,8 @@ class VIUpdateKwargs(PyBanditsBaseModel):
 
     num_steps: PositiveInt = 1000
     method: Literal["advi", "fullrank_advi"] = "advi"
-    optimizer_type: str = "sgd"
-    optimizer_kwargs: dict = Field(default_factory=lambda: {"step_size": 0.01})
+    optimizer_type: str = "adam"
+    optimizer_kwargs: dict = Field(default_factory=lambda: {"step_size": 0.0003})
     batch_size: Optional[PositiveInt] = None
     early_stopping_kwargs: Optional[dict] = None
     lr_scheduler_type: Optional[str] = None
--- a/pybandits/model/bnn/config.py
+++ b/pybandits/model/bnn/config.py
@@ -19,6 +19,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import warnings
 from copy import deepcopy
 from typing import ClassVar, List, Literal, Optional, Union
 
@@ -34,6 +35,7 @@ from pydantic import (
 from typing_extensions import Self
 
 from pybandits.base import (
+    PositiveFloat01,
     PyBanditsBaseModel,
 )
 from pybandits.model.bnn.priors import BaseLocationScaleArray, NormalArray, StudentTArray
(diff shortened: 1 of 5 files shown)


pybandits/model/bnn/config.py
```diff
@@ -280,3 +282,102 @@ def should_stop(self, loss: float) -> bool:
...
+        Fraction of total steps over which the KL term is linearly warmed up; must lie in
+        the half-open interval (0, 1] or be ``None``. Default None.
+    epochs : Optional[PositiveInt]
+        Number of epochs; takes precedence over ``num_steps`` when provided. Default None.
+    """
+
+    num_steps: PositiveInt = 1000
+    method: Literal["advi", "fullrank_advi"] = "advi"
+    optimizer_type: str = "sgd"
+    optimizer_kwargs: dict = Field(default_factory=lambda: {"step_size": 0.01})
```
ronshiff1: Do you think SGD + this LR are good defaults?

pybandits/model/bnn/network.py
```diff
@@ -542,54 +508,34 @@ def hidden_dim_list(self) -> List[int]:
...
-            for param in self._vi_update_params:
-                if param in self.update_kwargs:
-                    raise ValueError(
-                        f"Invalid update MCMC parameter: {param}. {self._vi_update_params} are VI parameters."
-                    )
-
-            self.update_kwargs = {**self._default_mcmc_kwargs, **self.update_kwargs}
+    @model_validator(mode="before")
+    @classmethod
+    def _coerce_update_kwargs(cls, data):
```
ronshiff1: why no typehints? and improve docstring (no description what's data)

pybandits/model/bnn/network.py
```diff
@@ -542,54 +508,34 @@ def hidden_dim_list(self) -> List[int]:
...
+        raw = data.get("update_kwargs")
+        if isinstance(raw, (VIUpdateKwargs, MCMCUpdateKwargs)):
+            return data
+        update_method = data.get("update_method", cls.model_fields["update_method"].default)
+        raw = dict(raw) if raw else {}
+        # Copy before mutating: during Union resolution pydantic feeds the same input dict to each
+        # candidate member, so mutating it here would leak an injected update_kwargs into sibling
+        # models (e.g. QuantitativeBayesianNeuralNetwork, which has no such field).
+        data = {**data}
+        if update_method == "VI":
```
ronshiff1, line 539: should we wrap this init with try catch?
(or we provide only the  pydantic error)

ronshiff1 (changes requested):

pybandits/model/bnn/network.py
```diff
@@ -542,54 +508,34 @@ def hidden_dim_list(self) -> List[int]:
...
+
+        Accepts a raw dict (the public API, and the form stored in serialized state), ``None``
+        (defaults), or an already-built ``VIUpdateKwargs``/``MCMCUpdateKwargs``. ``update_method``
+        is the single source of truth for which schema applies, so the kwargs themselves carry no
+        discriminator. All value validation and defaulting lives in the two typed models.
+        """
+        if not isinstance(data, dict):
+            return data
+        raw = data.get("update_kwargs")
+        if isinstance(raw, (VIUpdateKwargs, MCMCUpdateKwargs)):
```
ronshiff1, line 525: I agree with the bot's review comment - when isinstance(raw, (VIUpdateKwargs, MCMCUpdateKwargs) = True - we dont validate it matches the update_method

pybandits/model/bnn/network.py
shaharbar1: The pydantic error is more precise.

pybandits/model/bnn/config.py
shaharbar1: Fixed

pybandits/model/bnn/network.py
shaharbar1: Fixed

pybandits/model/bnn/network.py
shaharbar1: We don't have typehints for these validators almost anywhere in the repo

shaharbar1: Anyhow, I added

ronshiff1 (approved):

PR #154, opened 2026-07-02
Add limited actions exploration throttle

shaharbar1:
### Changes:

* Add limited_actions and limited_action_fraction fields to BaseMab — throttle over-exploration of newly-introduced arms
  - limited_actions competes only with probability limited_action_fraction on each selection, otherwise masked out
* Update BaseMab.__init__ and cold_start() to accept and propagate the two new params
* Add validation in the model validator — both must be set together, limited_actions must be a subset of the action set and must not contain default_action
* Update _predict_actions selection path to mask limited_actions via self._rng.binomial(1, limited_action_fraction)
  - skips masking when it would empty the action set

## Summary by CodeRabbit

* **New Features**
  * Added epsilon-greedy action limiting via `limited_actions` and `limited_action_fraction`, including probabilistic masking during exploration.
  * Extended CMAB/sMAB configurations with `default_action_fraction` and `limited_action_fraction`.
* **Bug Fixes**
  * Added validation to require both limiting parameters together, ensure `limited_actions` is a subset of available actions, and prevent limiting the default action.
  * Improved action-id handling to reject ids that are declared as both regular and quantitative.
* **Tests**
  * Added coverage for validation, masking vs competition, and empirical selection-share behavior.
  * Improved test determinism by switching many random inputs to use injected RNGs.
* **Chores**
  * Bumped package version to **7.5.1**.

### DIFF
--- a/pybandits/mab.py
+++ b/pybandits/mab.py
@@ -105,6 +105,14 @@ class BaseMab(PyBanditsBaseModel, ABC):
             - `default_action_fraction = 1.0`: explore always returns `default_action` (same as omitting this field).
             - `None` (default): legacy behavior; explore returns `default_action` deterministically when set,
               otherwise a uniformly-random action.
+    limited_actions : Optional[Set[ActionId]], None if not specified.
+        A set of actions whose selection is throttled, e.g. newly-introduced arms that Thompson Sampling
+        would otherwise over-explore. On each selection they are allowed to compete only with probability
+        `limited_action_fraction`; otherwise they are masked out. Must be used together with
+        `limited_action_fraction`, must be a subset of the action set, and must not contain `default_action`.
+    limited_action_fraction : Optional[Float01], None if not specified.
+        Probability that the `limited_actions` are allowed to compete on a given selection (mirrors
+        `epsilon`: higher = more exploration). Only meaningful together with `limited_actions`.
     current_supported_version_th : ClassVar[str]
         The threshold of the supported version of PyBandits which don't require any changes to the state.
     strategy_kwargs : Dict[str, Any]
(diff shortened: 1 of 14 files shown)


pybandits/mab.py
```diff
@@ -104,6 +104,14 @@ class BaseMab(PyBanditsBaseModel, ABC):
...
             - `None` (default): legacy behavior; explore returns `default_action` deterministically when set,
               otherwise a uniformly-random action.
+    limited_actions : Optional[Set[ActionId]], None if not specified.
+        A set of actions whose selection is throttled, e.g. newly-introduced arms that Thompson Sampling
+        would otherwise over-explore. On each selection they are allowed to compete only with probability
+        `limited_action_fraction`; otherwise they are masked out. Must be used together with
+        `limited_action_fraction`, must be a subset of the action set, and must not contain `default_action`.
+    limited_action_fraction : Optional[Float01], None if not specified.
+        Probability that the `limited_actions` are allowed to compete on a given selection (mirrors
+        `epsilon`: higher = more exploration). Only meaningful together with `limited_actions`.
```
AnastasiiaKabeshova, line 115: a bit misleading
can be read as:  higher = more **overall** exploration
Worth rewording to smth like: higher fraction  =  limited arms compete more often...

pybandits/mab.py
```diff
@@ -116,6 +124,8 @@ class BaseMab(PyBanditsBaseModel, ABC):
     epsilon: Optional[Float01] = None
     default_action: Optional[UnifiedActionId] = None
     default_action_fraction: Optional[PositiveFloat01] = None
+    limited_actions: Optional[Set[ActionId]] = None
+    limited_action_fraction: Optional[Float01] = None
```
AnastasiiaKabeshova, line 129: Is it the same fraction for all limited_actions?
should it also be a Set?

AnastasiiaKabeshova (approved):

PR #155, opened 2026-07-22
Add backbone_anchor_strength to CmabMetaModel

AnastasiiaKabeshova:
### Changes:

* Add backbone_anchor_strength field to CmabMetaModel — L2 penalty tying each update()'s backbone weights/biases to their pre-call values, the point-estimate analogue of the heads' own KL-to-previous-posterior anchor
  - defaults to 0.0 (disabled), reproducing the prior unanchored behavior exactly
  - addresses unbounded backbone drift under continual small-batch update() calls (e.g. hourly streaming retraining)
* Update CmabMetaModel.__init__() — thread backbone_anchor_strength through the kwargs bag the same way as backbone/random_seed (direct arg or cold-start kwargs)
* Update _declare_backbone() — add the numpyro.factor anchor term when backbone_anchor_strength > 0; applies on both the single-objective/minibatch path (_so_model) and the MO/quantitative full-batch path (_full_batch_model), since both call _declare_backbone
* Add backbone_anchor_strength test coverage to TestCmabMetaModel in tests/test_cmab_meta_model.py
  - zero strength matches default unanchored behavior exactly
  - positive strength measurably reduces backbone drift vs. unanchored, same seed/data
  - serializes/deserializes correctly via model_dump_json/model_validate_json
  - reaches the model via the cold-start kwargs bag (backbone_hidden_dims path), not just the direct constructor arg
  - is a silent no-op when backbone=None, even with a nonzero strength
* Update pyproject.toml — 8.0.0 -> 8.1.0, minor bump for the new backward-compatible field

## Summary by CodeRabbit

* **New Features**
  * Added joint shared-backbone training controls: `l2_anchoring` (weight/bias anchoring penalty) and `lr` (independent backbone learning rate).
  * Supports `lr=0.0` to freeze backbone updates while heads continue training.
  * Exposed a public `obj_optimizer` property on Bayesian neural network instances.
* **Bug Fixes**
  * Backbone-only knobs are now validated and error if used without enabling the shared-backbone path.
* **Tests**
  * Expanded coverage for anchoring drift reduction, backbone freezing, knob persistence, and JSON/kwargs propagation.
* **Chores**
  * Bumped package version to 8.1.0.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,6 +27,9 @@ repos:
     rev: 0.8.1
     hooks:
       - id: nbstripout
+        # fastjsonschema>=2.22.0 uses `dict | bool` PEP 604 syntax without a `__future__` import,
+        # which breaks the CI Python 3.9 job; pin below that until 3.9 support is dropped.
+        additional_dependencies: ['fastjsonschema<2.22.0']
         args: [
           '--extra-keys',
           'metadata.kernelspec metadata.language_info.version metadata.language_info.pygments_lexer'
--- a/pybandits/meta_model/cmab_meta_model.py
+++ b/pybandits/meta_model/cmab_meta_model.py
@@ -41,6 +41,7 @@ import jax
 import jax.numpy as jnp
 import numpy as np
 import numpyro
+import optax
 from loguru import logger
 from numpyro.distributions import Bernoulli as NumpyroBernoulli
 from numpyro.infer import TraceMeanField_ELBO
@@ -92,6 +93,9 @@ class CmabMetaModel(BaseMetaModel, Generic[CmabHeadType]):
     * a single BNN (SO/CC/DP);
     * a multi-objective head — a list of per-objective BNNs (``BaseBayesianNeuralNetworkMO``);
     * a quantitative head — wraps a ``.bnn`` whose input is ``[quantity ‖ context]``.
+
+    How the shared encoder is trained (anchoring penalty, backbone-only learning rate) is configured on
+    the backbone itself — see :class:`~pybandits.model.bnn.backbone.MLPBackbone`.
     """
 
     actions: Dict[ActionId, CmabHeadType]  # type: ignore[assignment]
@@ -106,6 +110,9 @@ class CmabMetaModel(BaseMetaModel, Generic[CmabHeadType]):
     _obj_prefix: ClassVar[str] = "obj"
     # tqdm progress-bar label for the joint SVI run.
     _svi_desc: ClassVar[str] = "cmab joint SVI"
+    # Cold-start kwargs prefix marking a key as backbone-only: it is popped and forwarded, prefix
+    # stripped, to ``MLPBackbone.cold_start`` (so a new backbone knob needs no change here).
+    _backbone_kwargs_prefix: ClassVar[str] = "backbone_"
 
     _rng_key: Any = PrivateAttr(default=None)
 
@@ -121,11 +128,12 @@ class CmabMetaModel(BaseMetaModel, Generic[CmabHeadType]):
         """Build from a pre-made ``actions`` dict or cold-start specs, plus an optional ``backbone``.
 
         The shared ``backbone`` / ``random_seed`` fields reach this constructor two ways: inside the
-        ``kwargs`` bag on **cold start** (the backbone requested via ``backbone_hidden_dims`` plus
-        optional ``embedding_dim`` / ``backbone_activation`` — how ``cold_start`` reaches this class
-        through the generic ``BaseMab.cold_start`` factory, so no bespoke cmab ``cold_start`` is needed),
-        or as top-level keyword arguments on **pydantic (de)serialization**. Both are folded into the
-        single bag here so ``BaseMetaModel.__init__`` handles every field uniformly (no catch-all kwarg).
+        ``kwargs`` bag on **cold start** (the backbone requested via ``backbone_hidden_dims`` plus any
+        other ``backbone_``-prefixed knob — see ``_build_backbone_from_kwargs``; this is how
+        ``cold_start`` reaches this class through the generic ``BaseMab.cold_start`` factory, so no
+        bespoke cmab ``cold_start`` is needed), or as top-level keyword arguments on **pydantic
+        (de)serialization**. Both are folded into the single bag here so ``BaseMetaModel.__init__``
+        handles every field uniformly (no catch-all kwarg).
         """
         kwargs = kwargs or {}
         # Fold the explicitly-passed (de)serialization fields into the bag; ``setdefault`` yields to a
@@ -138,7 +146,7 @@ class CmabMetaModel(BaseMetaModel, Generic[CmabHeadType]):
         # heads that ``BaseMetaModel.__init__`` cold-starts sit on the embedding. The built backbone is
         # stashed back into the bag as the ``backbone`` field for the base ``__init__`` to forward.
         if kwargs.get("backbone") is None:
-            built = self._build_backbone_from_kwargs(kwargs, kwargs.get("random_seed"))
+            built = self._build_backbone_from_kwargs(kwargs)
             if built is not None:
                 kwargs["backbone"] = built
         super().__init__(
@@ -149,41 +157,34 @@ class CmabMetaModel(BaseMetaModel, Generic[CmabHeadType]):
         )
 
     @classmethod
-    def _build_backbone_from_kwargs(
-        cls, kwargs: Dict[str, Any], random_seed: Optional[NonNegativeInt]
-    ) -> Optional[MLPBackbone]:
-        """Pop the backbone specs from ``kwargs`` and build the shared encoder (``None`` if not requested).
+    def _build_backbone_from_kwargs(cls, kwargs: Dict[str, Any]) -> Optional[MLPBackbone]:
+        """Pop every ``backbone_``-prefixed key from ``kwargs`` and build the shared encoder (``None``
+        if not requested).
 
-        Mutates ``kwargs`` in place: removes ``backbone_hidden_dims``/``embedding_dim``/
-        ``backbone_activation`` and, when a backbone is built, rewrites ``n_features`` to the embedding
-        width so the per-arm heads sit on the embedding. Backbone-only knobs without
-        ``backbone_hidden_dims`` raise, rather than being silently ignored.
+        Mutates ``kwargs`` in place: each ``f"{_backbone_kwargs_prefix}{name}"`` key is popped and,
+        prefix stripped, forwarded as ``name=`` to :meth:`MLPBackbone.cold_start` — so a new
+        backbone-only knob needs a ``cold_start`` parameter there and nothing here. Keys not given keep
+        ``MLPBackbone.cold_start``'s own defaults rather than duplicating them here, except
+        ``random_seed``, which falls back to the meta-model's own so one top-level seed makes the whole
+        model reproducible. When a backbone is built, ``n_features`` is rewritten to the embedding width
+        so the per-arm heads sit on the embedding. Backbone-only knobs without ``hidden_dims`` (i.e. no
+        backbone requested) raise, rather than being silently ignored.
         """
-        hidden_dims = kwargs.pop("backbone_hidden_dims", None)
-        embedding_dim = kwargs.pop("embedding_dim", None)
-        activation = kwargs.pop("backbone_activation", None)
-        if hidden_dims is None:
-            backbone_only = {"embedding_dim": embedding_dim, "backbone_activation": activation}
-            set_params = sorted(name for name, value in backbone_only.items() if value is not None)
-            if set_params:
+        prefix = cls._backbone_kwargs_prefix
+        backbone_kwargs = {key[len(prefix) :]: kwargs.pop(key) for key in list(kwargs) if key.startswith(prefix)}
+        if "hidden_dims" not in backbone_kwargs:
+            if backbone_kwargs:
+                set_params = sorted(f"{prefix}{name}" for name in backbone_kwargs)
                 raise TypeError(f"{set_params} only apply with a backbone; pass backbone_hidden_dims=[...].")
             return None
         n_features = kwargs.get("n_features")
         if n_features is None:
             raise ValueError("n_features must be provided to build the shared backbone.")
-        emb = (
-            embedding_dim
-            if embedding_dim is not None
-            else max(1, n_features // BaseBayesianNeuralNetwork.embedding_dim_divisor)
-        )
-        backbone = MLPBackbone.cold_start(
-            n_features=n_features,
-            hidden_dims=hidden_dims,
-            embedding_dim=emb,
-            activation=activation or "relu",
-            random_seed=kwargs.get("random_seed", random_seed),
-        )
-        kwargs["n_features"] = emb  # per-arm heads live on the embedding, not the raw context
+        backbone_kwargs.setdefault("random_seed", kwargs.get("random_seed"))
+        if backbone_kwargs.get("embedding_dim") is None:
+            backbone_kwargs["embedding_dim"] = max(1, n_features // BaseBayesianNeuralNetwork.embedding_dim_divisor)
+        backbone = MLPBackbone.cold_start(n_features=n_features, **backbone_kwargs)
+        kwargs["n_features"] = backbone_kwargs["embedding_dim"]  # heads live on the embedding, not raw context
         return backbone
 
     def model_post_init(self, __context: Any) -> None:
@@ -448,7 +449,7 @@ class CmabMetaModel(BaseMetaModel, Generic[CmabHeadType]):
         svi, params, _history, self._rng_key = run_svi(
             model=model,
             guide=guide,
-            optimizer=representative_bnn._obj_optimizer,
+            optimizer=self._get_joint_optimizer(representative_bnn, n_bb_layers),
             loss=loss,
             rng_key=self._rng_key,
             model_args=model_args,
@@ -462,12 +463,56 @@ class CmabMetaModel(BaseMetaModel, Generic[CmabHeadType]):
             self._store_head(self.actions[arm], site_mu, site_sigma, arm)
         self._store_backbone(params, n_bb_layers)
 
+    def _get_joint_optimizer(self, representative_bnn: BaseBayesianNeuralNetwork, n_bb_layers: int) -> Any:
+        """The optimizer for the joint SVI pass.
+
+        One optimizer for every parameter (backbone and heads alike) unless the shared backbone sets its
+        own ``lr``, in which case an ``optax.multi_transform`` gives the backbone's params their own rate
+        — everything else (optimizer type, remaining ``optimizer_kwargs``, lr schedule) stays shared with
+        the heads. Gradient clipping is applied globally, *before* the per-group split, so the threshold
+        keeps bounding the whole gradient's norm either way.
+
+        ``backbone.lr=0.0`` uses ``optax.set_to_zero()`` rather than the adaptive optimizer at a zero
+        rate, so the per-step update is architecturally zero whatever the optimizer's internals do.
+        """
+        if self.backbone is None or self.backbone.lr is None:
+            return representative_bnn.obj_optimizer
+
+        backbone_optimizer = (
+            optax.set_to_zero()
+            if not self.backbone.lr
+            else representative_bnn.build_optax_optimizer(step_size=self.backbone.lr)
+        )
+        backbone_param_names: Set[str] = {
+            name for i in range(n_bb_layers) for name in self.backbone.get_layer_params_name(i)
+        }
+
+        def label_fn(params: Dict[str, Any]) -> Dict[str, str]:
+            return {name: ("backbone" if name in backbone_param_names else "head") for name in params}
+
+        return representative_bnn.clip_and_wrap_optimizer(
+            optax.multi_transform(
+                {"backbone": backbone_optimizer, "head": representative_bnn.build_optax_optimizer()}, label_fn
+            )
+        )
+
     def _declare_backbone(self, backbone_w: list, backbone_b: list, n_bb_layers: int) -> list:
-        """Register the backbone's deterministic weights as ``numpyro.param`` (inside the trace)."""
+        """Register the backbone's deterministic weights as ``numpyro.param`` (inside the trace).
+
+        When ``backbone.l2_anchoring > 0``, also adds a ``numpyro.factor`` quadratic penalty pulling each
+        layer's optimized weights *and* biases back toward ``backbone_w``/``backbone_b`` — the values at
+        the start of this SVI run, i.e. the previous ``update()`` call's — to bound per-call drift (see
+        :class:`~pybandits.model.bnn.backbone.MLPBackbone`). Declared outside the data plate, so the
+        penalty is not scaled by the minibatch.
+        """
         weights_biases = []
         for i in range(n_bb_layers):
             w_name, b_name = self.backbone.get_layer_params_name(i)
-            weights_biases.append((numpyro.param(w_name, backbone_w[i]), numpyro.param(b_name, backbone_b[i])))
+            w, b = numpyro.param(w_name, backbone_w[i]), numpyro.param(b_name, backbone_b[i])
+            if self.backbone.l2_anchoring > 0:
+                penalty = jnp.sum((w - backbone_w[i]) ** 2) + jnp.sum((b - backbone_b[i]) ** 2)
+                numpyro.factor(f"backbone_anchor_{i}", -0.5 * self.backbone.l2_anchoring * penalty)
+            weights_biases.append((w, b))
         return weights_biases
 
     def _full_batch_model(
--- a/pybandits/model/bnn/backbone.py
+++ b/pybandits/model/bnn/backbone.py
@@ -33,7 +33,7 @@ from typing import ClassVar, List, Optional, Tuple
 
 import jax.numpy as jnp
 import numpy as np
-from pydantic import ConfigDict, NonNegativeInt, PositiveInt
+from pydantic import ConfigDict, NonNegativeFloat, NonNegativeInt, PositiveInt
 from typing_extensions import Self
 
 from pybandits.model.bnn._dnn import DNNMixin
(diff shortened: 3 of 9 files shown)


pybandits/meta_model/cmab_meta_model.py
```diff
@@ -97,6 +97,21 @@ class CmabMetaModel(BaseMetaModel, Generic[CmabHeadType]):
     actions: Dict[ActionId, CmabHeadType]  # type: ignore[assignment]
     backbone: Optional[MLPBackbone] = None
     random_seed: Optional[NonNegativeInt] = None
+    # Strength of an L2 penalty (a numpyro.factor term) tying each update()'s backbone weights/biases
```
shaharbar1: This comment location is odd, put it on top with rest of the docstring

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -97,6 +97,21 @@ class CmabMetaModel(BaseMetaModel, Generic[CmabHeadType]):
...
+    # unbounded per-call movement compounds and degrades online performance over time, even though
+    # each individual update looks fine in isolation.
+    #
+    # 0.0 (default) disables the penalty and reproduces the exact prior (unanchored) behavior — no
+    # backward-compatibility concerns. There is no universally-good nonzero default: the right
+    # magnitude trades off against the likelihood term, which scales with batch size and network
+    # size, so it must be picked per deployment (start small, e.g. 1e3-1e5, and increase until
+    # per-call backbone drift stabilizes without stalling the backbone entirely). Only meaningful
+    # when a shared `backbone` is set; silently unused otherwise.
+    backbone_anchor_strength: float = 0.0
```
shaharbar1: Maybe backbone_anchoring_factor?

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -117,6 +132,7 @@ def __init__(
         kwargs: Optional[Dict[str, Any]] = None,
         backbone: Optional[MLPBackbone] = None,
         random_seed: Optional[NonNegativeInt] = None,
+        backbone_anchor_strength: Optional[float] = None,
```
shaharbar1: Please add validation that this parameter is provided only if backbone is provided. With matching tests of course 😉

pybandits/meta_model/cmab_meta_model.py
shaharbar1: Also shouldn't this be an attribute of mlpbackbone class?

shaharbar1 (changes requested):

shaharbar1:
In general, I didn't understand the motivation for anchoring... I agree regularization is needed for the backbone, but why not just l2 regularization?

shaharbar1:
> In general, I didn't understand the motivation for anchoring... I agree regularization is needed for the backbone, but why not just l2 regularization?

OK, there is this work for example 
https://www.pnas.org/doi/10.1073/pnas.1611835114
So now I understand better. So most of my comments are irrelevant because it really should be anchoring, and not weight decay. Good move 👏 @AnastasiiaKabeshova 

Just consider moving the new attribute into the mlp backbone class...

AnastasiiaKabeshova:
> > In general, I didn't understand the motivation for anchoring... I agree regularization is needed for the backbone, but why not just l2 regularization?
> 
> OK, there is this work for example https://www.pnas.org/doi/10.1073/pnas.1611835114 So now I understand better. So most of my comments are irrelevant because it really should be anchoring, and not weight decay. Good move 👏 @AnastasiiaKabeshova
> 
> Just consider moving the new attribute into the mlp backbone class...

Thanks a lot for checking PR! :) 

Yes, you are right. When doing experiments I observed:
- on hourly streaming updates, backbone drift stayed ~5-10% of weight norm per call indefinitely (never settling), and online AUC/PR-AUC degraded over 2-3 days vs. a steady no-backbone baseline.
-  freezing the backbone after pretraining (heads still streaming) fully restored stable performance 
-  Fix: an optional numpyro.factor L2 penalty tying each call's backbone weights to their pre-call values ‚Äî the same anchor the heads get implicitly, made explicit for the backbone.
- Default: 0.0 (off, fully backward compatible) - the right strength is deployment-specific (1e5 worked best on the real eval; 1e6+ over-constrains).

shaharbar1:
> > > In general, I didn't understand the motivation for anchoring... I agree regularization is needed for the backbone, but why not just l2 regularization?
> > 
> > 
> > OK, there is this work for example https://www.pnas.org/doi/10.1073/pnas.1611835114 So now I understand better. So most of my comments are irrelevant because it really should be anchoring, and not weight decay. Good move 👏 @AnastasiiaKabeshova
> > Just consider moving the new attribute into the mlp backbone class...
> 
> Thanks a lot for checking PR! :)
> 
> Yes, you are right. When doing experiments I observed:
> 
> * on hourly streaming updates, backbone drift stayed ~5-10% of weight norm per call indefinitely (never settling), and online AUC/PR-AUC degraded over 2-3 days vs. a steady no-backbone baseline.
> * freezing the backbone after pretraining (heads still streaming) fully restored stable performance
> * Fix: an optional numpyro.factor L2 penalty tying each call's backbone weights to their pre-call values ‚Äî the same anchor the heads get implicitly, made explicit for the backbone.
> * Default: 0.0 (off, fully backward compatible) - the right strength is deployment-specific (1e5 worked best on the real eval; 1e6+ over-constrains).

@AnastasiiaKabeshova This strength value sounds quite big, no?
Did you verify that with lr=0 the backbone remains the same?
Can it be that there is a bug somewhere like we had for the bnn in the past?

AnastasiiaKabeshova:
> > > > In general, I didn't understand the motivation for anchoring... I agree regularization is needed for the backbone, but why not just l2 regularization?
> > > 
> > > 
> > > OK, there is this work for example https://www.pnas.org/doi/10.1073/pnas.1611835114 So now I understand better. So most of my comments are irrelevant because it really should be anchoring, and not weight decay. Good move 👏 @AnastasiiaKabeshova
> > > Just consider moving the new attribute into the mlp backbone class...
> > 
> > 
> > Thanks a lot for checking PR! :)
> > Yes, you are right. When doing experiments I observed:
> > 
> > * on hourly streaming updates, backbone drift stayed ~5-10% of weight norm per call indefinitely (never settling), and online AUC/PR-AUC degraded over 2-3 days vs. a steady no-backbone baseline.
> > * freezing the backbone after pretraining (heads still streaming) fully restored stable performance
> > * Fix: an optional numpyro.factor L2 penalty tying each call's backbone weights to their pre-call values ‚Äî the same anchor the heads get implicitly, made explicit for the backbone.
> > * Default: 0.0 (off, fully backward compatible) - the right strength is deployment-specific (1e5 worked best on the real eval; 1e6+ over-constrains).
> 
> @AnastasiiaKabeshova This strength value sounds quite big, no? Did you verify that with lr=0 the backbone remains the same? Can it be that there is a bug somewhere like we had for the bnn in the past?

Yes, thanks
I checked backbone and head weight's drift while lr = 0 , it is not drifting

shaharbar1:
> > > > > In general, I didn't understand the motivation for anchoring... I agree regularization is needed for the backbone, but why not just l2 regularization?
> > > > 
> > > > 
> > > > OK, there is this work for example https://www.pnas.org/doi/10.1073/pnas.1611835114 So now I understand better. So most of my comments are irrelevant because it really should be anchoring, and not weight decay. Good move 👏 @AnastasiiaKabeshova
> > > > Just consider moving the new attribute into the mlp backbone class...
> > > 
> > > 
> > > Thanks a lot for checking PR! :)
> > > Yes, you are right. When doing experiments I observed:
> > > 
> > > * on hourly streaming updates, backbone drift stayed ~5-10% of weight norm per call indefinitely (never settling), and online AUC/PR-AUC degraded over 2-3 days vs. a steady no-backbone baseline.
> > > * freezing the backbone after pretraining (heads still streaming) fully restored stable performance
> > > * Fix: an optional numpyro.factor L2 penalty tying each call's backbone weights to their pre-call values ‚Äî the same anchor the heads get implicitly, made explicit for the backbone.
> > > * Default: 0.0 (off, fully backward compatible) - the right strength is deployment-specific (1e5 worked best on the real eval; 1e6+ over-constrains).
> > 
> > 
> > @AnastasiiaKabeshova This strength value sounds quite big, no? Did you verify that with lr=0 the backbone remains the same? Can it be that there is a bug somewhere like we had for the bnn in the past?
> 
> Yes, thanks I checked backbone and head weight's drift while lr = 0 , it is not drifting

@AnastasiiaKabeshova you could consider also a dedicated lr for the backbone, maybe even to freeze it after pretraining.
Anyhow, nice idea and implementation. Ping me with a tag when you need approval.

shaharbar1 (changes requested):
My main comment: the new attributes should be put into the mlp backbone class, rather than the cmab meta model.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -155,15 +165,22 @@ def _build_backbone_from_kwargs(
...
+        Backbone-only knobs without ``backbone_hidden_dims`` raise, rather than being silently ignored.
         """
         hidden_dims = kwargs.pop("backbone_hidden_dims", None)
         embedding_dim = kwargs.pop("embedding_dim", None)
         activation = kwargs.pop("backbone_activation", None)
+        l2_reg_factor = kwargs.pop("backbone_l2_reg_factor", None)
+        backbone_lr = kwargs.pop("backbone_lr", None)
         if hidden_dims is None:
-            backbone_only = {"embedding_dim": embedding_dim, "backbone_activation": activation}
+            backbone_only = {
```
shaharbar1: Instead of this, consider adding a const like we have action_ids_prefix with the value "backbone" and use it to automatically pass all parameters with this prefix to the backbone model. For this, you need to change embedding_dim to backbone_embedding_dim. Your call.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -182,6 +199,8 @@ def _build_backbone_from_kwargs(
             embedding_dim=emb,
             activation=activation or "relu",
             random_seed=kwargs.get("random_seed", random_seed),
+            backbone_l2_reg_factor=l2_reg_factor if l2_reg_factor is not None else 0.0,
```
shaharbar1: The none replacing should take place inside the backbone class, no?

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -462,12 +481,88 @@ def _joint_svi_update(
             self._store_head(self.actions[arm], site_mu, site_sigma, arm)
         self._store_backbone(params, n_bb_layers)
 
+    def _get_joint_optimizer(self, representative_bnn: BaseBayesianNeuralNetwork, n_bb_layers: int) -> Any:
+        """The optimizer for the joint SVI pass — one optimizer for every parameter (backbone and
+        heads alike) by default, exactly as before `backbone_lr` existed; a per-parameter-group
```
shaharbar1: Too much details with the "before backbone_lr" existed. This is too llm-ish.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -462,12 +481,88 @@ def _joint_svi_update(
...
+
+        Note this does not, by itself, guarantee the backbone's stored weights are bit-identical at
+        float64 precision across an ``update()`` call: the backbone stores ``float64`` weights, but
+        every joint-SVI computation runs through JAX at its default ``float32`` precision, so any value
+        that passes through the pipeline — frozen or not — gets re-quantized to float32 precision on the
+        way back out (a pre-existing property of this class, unrelated to `backbone_lr`). "Exactly
+        frozen" should be read/tested at that float32 working precision, not float64.
+        """
+        if self.backbone is None or self.backbone.backbone_lr is None:
+            return representative_bnn._obj_optimizer
```
shaharbar1: Make obj_optimizer unprotected.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -462,12 +481,88 @@ def _joint_svi_update(
...
+            if update_kwargs.lr_scheduler_type is None:
+                return init_value
+            scheduler_fn = representative_bnn._resolve_optax_fn(update_kwargs.lr_scheduler_type, "lr_scheduler")
+            return scheduler_fn(init_value=init_value, **(update_kwargs.lr_scheduler_kwargs or {}))
+
+        try:
+            head_optimizer = optimizer_fn(learning_rate=_resolve_lr(head_lr), **optimizer_kwargs)
+            backbone_optimizer = (
+                optax.set_to_zero()
+                if backbone_lr == 0.0
```
shaharbar1: You can test for not backbone_lr rather than comparing to zero. Not sure which is better, but it will cover none case easily.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -462,12 +481,88 @@ def _joint_svi_update(
...
+        if update_kwargs.gradient_clip_norm is not None:
+            multi_optimizer = optax.chain(optax.clip_by_global_norm(update_kwargs.gradient_clip_norm), multi_optimizer)
+        return noptim.optax_to_numpyro(multi_optimizer)
+
     def _declare_backbone(self, backbone_w: list, backbone_b: list, n_bb_layers: int) -> list:
-        """Register the backbone's deterministic weights as ``numpyro.param`` (inside the trace)."""
+        """Register the backbone's deterministic weights as ``numpyro.param`` (inside the trace).
+
+        When ``backbone.backbone_l2_reg_factor > 0``, also adds a ``numpyro.factor`` L2 penalty tying
+        each layer's new *weights* (not biases, matching standard L2/weight-decay practice) to their
```
shaharbar1: I know this is due to my earlier comments, but I understand that this is not l2 reg, but anchoring, so you should stick to your earlier terminology of anchoring, rather than my suggest l2_reg. Possibly best suggestion is l2_anchoring.

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -462,12 +481,88 @@ def _joint_svi_update(
...
+        multi_optimizer = optax.multi_transform({"backbone": backbone_optimizer, "head": head_optimizer}, label_fn)
+        if update_kwargs.gradient_clip_norm is not None:
+            multi_optimizer = optax.chain(optax.clip_by_global_norm(update_kwargs.gradient_clip_norm), multi_optimizer)
+        return noptim.optax_to_numpyro(multi_optimizer)
+
     def _declare_backbone(self, backbone_w: list, backbone_b: list, n_bb_layers: int) -> list:
-        """Register the backbone's deterministic weights as ``numpyro.param`` (inside the trace)."""
+        """Register the backbone's deterministic weights as ``numpyro.param`` (inside the trace).
+
+        When ``backbone.backbone_l2_reg_factor > 0``, also adds a ``numpyro.factor`` L2 penalty tying
```
shaharbar1: You should take also the biases into the anchoring, this is not standard l2 reg. This is what I saw in the PNAS paper I mentioned before.

pybandits/model/bnn/backbone.py
```diff
@@ -65,6 +97,8 @@ class MLPBackbone(DNNMixin):
     random_seed: Optional[NonNegativeInt] = None
     weights: List[List[List[float]]]  # per layer: (input_dim, output_dim)
     biases: List[List[float]]  # per layer: (output_dim,)
+    backbone_l2_reg_factor: float = 0.0
+    backbone_lr: Optional[float] = None
```
shaharbar1: Should be nonnegative float

pybandits/model/bnn/backbone.py
```diff
@@ -65,6 +97,8 @@ class MLPBackbone(DNNMixin):
     random_seed: Optional[NonNegativeInt] = None
     weights: List[List[List[float]]]  # per layer: (input_dim, output_dim)
     biases: List[List[float]]  # per layer: (output_dim,)
+    backbone_l2_reg_factor: float = 0.0
```
shaharbar1: Non negative float

tests/test_cmab_meta_model.py
```diff
@@ -78,6 +80,27 @@ def test_serialization_round_trip(self, activation: str, rng: np.random.Generato
         x = self._context(rng)
         np.testing.assert_allclose(bb.embed(x), restored.embed(x))
 
+    def test_training_knobs_survive_reset_and_serialization(self) -> None:
```
shaharbar1: This test should utilize st.given rather than relying on class constants

tests/test_cmab_meta_model.py
```diff
@@ -228,6 +275,115 @@ def test_backbone_update_changes_backbone_and_heads(
...
+        unanchored.update(actions=actions, rewards=rewards, context=context)
+        anchored.update(actions=actions, rewards=rewards, context=context)
+        drift_unanchored = sum(np.sum((a - b) ** 2) for a, b in zip(w_initial, unanchored.backbone.weight_arrays))
+        drift_anchored = sum(np.sum((a - b) ** 2) for a, b in zip(w_initial, anchored.backbone.weight_arrays))
+        assert drift_anchored < drift_unanchored
+
+    def test_backbone_lr_zero_freezes_backbone_exactly(self, rng: np.random.Generator) -> None:
+        """``backbone_lr=0.0`` produces exactly zero backbone movement, while heads still train normally.
+
+        Compared at ``float32`` — the backbone stores ``float64`` weights, but every joint-SVI
```
shaharbar1: Why backbone uses float64? You can use float32 all the way.

tests/test_cmab_meta_model.py
```diff
@@ -228,6 +275,115 @@ def test_backbone_update_changes_backbone_and_heads(
...
+        "knob_kwarg",
+        [{"backbone_l2_reg_factor": l2_reg_factor_cold_start_value}, {"backbone_lr": backbone_lr_cold_start_value}],
+    )
+    def test_backbone_training_knobs_reach_backbone_via_cold_start_kwargs(self, knob_kwarg: dict) -> None:
+        """Backbone training knobs thread through the cold-start ``kwargs`` bag onto the built ``MLPBackbone``."""
+        kwargs = {
+            "n_features": self.n_features,
+            "hidden_dim_list": self.hidden,
+            "update_kwargs": {"num_steps": self.num_steps},
+            "backbone_hidden_dims": [8],
```
shaharbar1: Put 8 in test argument

.pre-commit-config.yaml
```diff
@@ -27,6 +27,9 @@ repos:
     rev: 0.8.1
     hooks:
       - id: nbstripout
+        # fastjsonschema>=2.22.0 uses `dict | bool` PEP 604 syntax without a `__future__` import,
+        # which breaks the CI Python 3.9 job; pin below that until 3.9 support is dropped.
+        additional_dependencies: ['fastjsonschema<2.22.0']
```
shaharbar1, line 32: Didn't mles move the docker to py3.12 yet?

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -462,12 +481,88 @@ def _joint_svi_update(
...
+        thing (a bound on the *whole* gradient's norm) whether or not `backbone_lr` is set.
+
+        `backbone_lr=0.0` uses ``optax.set_to_zero()`` for the backbone group rather than the adaptive
+        optimizer at a zero learning rate: this makes the per-step update architecturally always the
+        zero vector (not merely a value that evaluates to zero after being scaled by a 0.0 learning
+        rate), which is the more robust choice regardless of the optimizer type/internals.
+
+        Note this does not, by itself, guarantee the backbone's stored weights are bit-identical at
+        float64 precision across an ``update()`` call: the backbone stores ``float64`` weights, but
+        every joint-SVI computation runs through JAX at its default ``float32`` precision, so any value
```
shaharbar1: So why not stick with float32 all the way?

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -155,15 +165,22 @@ def _build_backbone_from_kwargs(
...
+        l2_reg_factor = kwargs.pop("backbone_l2_reg_factor", None)
+        backbone_lr = kwargs.pop("backbone_lr", None)
         if hidden_dims is None:
-            backbone_only = {"embedding_dim": embedding_dim, "backbone_activation": activation}
+            backbone_only = {
+                "embedding_dim": embedding_dim,
+                "backbone_activation": activation,
+                "backbone_l2_reg_factor": l2_reg_factor,
+                "backbone_lr": backbone_lr,
+            }
```
shaharbar1: In general, I wonder if you could make things more dynamic here, rather than defining all backbone arguments, use **kwargs and grab what the backbone needs, like you did somewhere else in the code with the utils.py arguments' grabbing function

shaharbar1 (changes requested):

pybandits/meta_model/cmab_meta_model.py
AnastasiiaKabeshova: Very good suggestion, thanks!

pybandits/meta_model/cmab_meta_model.py
```diff
@@ -152,38 +169,38 @@ def __init__(
...
         )
-        backbone = MLPBackbone.cold_start(
-            n_features=n_features,
-            hidden_dims=hidden_dims,
-            embedding_dim=emb,
-            activation=activation or "relu",
-            random_seed=kwargs.get("random_seed", random_seed),
-        )
-        kwargs["n_features"] = emb  # per-arm heads live on the embedding, not the raw context
+        backbone_kwargs.setdefault("activation", "relu")
```
shaharbar1: Why do you need this default? The default already exists on dnnmixin... I think the activation and random seed should be skipped in terms of setting default. Especially for activation, it causes hard coding of relu...

pybandits/model/bnn/backbone.py
```diff
@@ -65,20 +103,30 @@ class MLPBackbone(DNNMixin):
...
     weight_var_name: ClassVar[str] = "backbone_weight"
     bias_var_name: ClassVar[str] = "backbone_bias"
 
     @property
     def weight_arrays(self) -> List[np.ndarray]:
-        """Per-layer weight matrices as numpy arrays (consumed by the forward pass / joint SVI engine)."""
-        return [np.asarray(w, dtype=float) for w in self.weights]
+        """Per-layer weight matrices as numpy arrays (consumed by the forward pass / joint SVI engine).
+
+        ``float32``, matching the precision every joint-SVI/JAX computation actually runs at (JAX's
```
shaharbar1: Too much details in this docstring I think...

pybandits/model/bnn/backbone.py
```diff
@@ -65,20 +103,30 @@ class MLPBackbone(DNNMixin):
...
+        default) — reading these back at ``float64`` would silently re-introduce precision the training
+        pipeline never had in the first place.
+        """
+        return [np.asarray(w, dtype=np.float32) for w in self.weights]
```
shaharbar1: Add a dedicated static method for this conversion and comprehension as it is used repeatedly on weights and biases...

pybandits/model/bnn/backbone.py
```diff
@@ -190,6 +240,10 @@ def cold_start(
             Element-wise activation applied after every layer except the last.
         random_seed : Optional[NonNegativeInt], default=None
             Seed for reproducible initialisation.
+        l2_anchoring : NonNegativeFloat, default=0.0
+            See the class docstring's "Joint-training knobs" section.
+        lr : Optional[NonNegativeFloat], default=None
+            See the class docstring's "Joint-training knobs" section.
```
shaharbar1: I think this is overdocumentation

pybandits/model/bnn/network.py
```diff
@@ -262,6 +262,11 @@ def get_embedding_var_name(cls, feat_index: int) -> str:
     def approx_history(self) -> Optional[np.ndarray]:
         return self._approx_history
 
+    @property
+    def obj_optimizer(self) -> Any:
```
shaharbar1, line 266: Cool!

tests/test_cmab.py
```diff
@@ -1602,8 +1602,10 @@ def test_cold_start_with_backbone_reports_raw_input_dim() -> None:
     "kwargs, expected_exception, match",
     [
         # Backbone-only knobs are rejected when no backbone is requested.
-        (dict(embedding_dim=8), TypeError, "only apply with a backbone"),
+        (dict(backbone_embedding_dim=8), TypeError, "only apply with a backbone"),
         (dict(backbone_activation="tanh"), TypeError, "only apply with a backbone"),
```
shaharbar1: Better use st.given for running through multiple values...

tests/test_cmab_meta_model.py
```diff
@@ -122,9 +170,11 @@ def _build_meta(self, with_backbone: bool, batch_size: Optional[int] = None) ->
         if with_backbone:
             kwargs["backbone"] = MLPBackbone.cold_start(
                 n_features=self.n_features,
-                hidden_dims=[8],
+                hidden_dims=self.backbone_hidden_dims,
```
shaharbar1, line 173: Better use st.given for feeding values to the test, rather than relying on consts...

tests/test_cmab_meta_model.py
```diff
@@ -228,6 +278,105 @@ def test_backbone_update_changes_backbone_and_heads(
...
+        """
+        kwargs = {
+            "n_features": self.n_features,
+            "hidden_dim_list": self.hidden,
+            "update_kwargs": {"num_steps": self.num_steps},
+            f"backbone_{knob_name}": value,
+        }
+        with pytest.raises(TypeError, match="only apply with a backbone"):
+            CmabMetaModelSO(action_ids=self.valid_subset, kwargs=kwargs)
+
```
shaharbar1, line 424: Very nice tests that you added 👏

.pre-commit-config.yaml
shaharbar1: @AnastasiiaKabeshova if they did, then just remove py3.9 support in toml and on ci cd

shaharbar1 (approved):
@AnastasiiaKabeshova approved! Well done! Just mind my final comments! Good luck!

.pre-commit-config.yaml
AnastasiiaKabeshova: @shaharbar1  No, it is not done yet
When it is done, I will do another pr :)

shaharbar1:
@AnastasiiaKabeshova please note that the previous NLB PR made the ci/cd pytest extremely slow, like 2 hours instead of 8 minutes. See e.g. here https://github.com/PlaytikaOSS/pybandits/actions/runs/29036940647/job/86184126835.
The test durations at the end of the ci step show that test_update tests are really long now, with mocc test taking 30 minutes. Maybe the mocks are not really working since the introduction of the NLB as they worked on mocking the bnn updates directly I think, but those are not relevant. So I guess the mocks should be redirected through the test suite.

PR #156, opened 2026-07-28
Add model execution-time benchmark script and workflow

### LINKED ISSUE #78: Model Benchmarking script
Add a script that runs all available sMABs and cMABs, and benchmarks their "update" and "predict" execution time for various sample sizes: 100, 1000, 10000, 100000.

Whatsonyourmind:
### Changes:
* Adds `benchmarks/benchmark_models.py`, timing `update` and `predict` for every sMAB and cMAB across a range of sample sizes and writing one CSV row per measurement (closes #78). sMABs and cMABs are benchmarked separately, and the cMABs are timed in two passes — a fixed step budget for comparable per-step cost, and an `epochs` pass with an explicit `batch_size` for the realistic cost, reporting the steps actually taken.
* Adds `.github/workflows/benchmark.yml`, running the reduced sweep on pull requests and the full sweep on manual dispatch, uploading the CSV as an artifact.
* Tests added: `tests/test_benchmark_models.py`, 14 cases covering registry completeness, reward shaping, both pass configurations, repeat isolation, the zero-repeats guard, per-action step accounting, and one end-to-end sMAB sweep. Runs in ~5s; the cMAB path is deliberately excluded, since one SVI update takes seconds and would turn the suite into the benchmark.
* Documentation: the design decisions are documented in the module docstring — why the families are separated, why two passes, why a warm-up precedes every measurement, and why `epochs` needs an explicit `batch_size` to scale steps with the data.
* Minor version bump to `8.1.0`.

## Summary by CodeRabbit

- **New Features**
  - Added a benchmarking command-line tool covering all supported sMAB and cMAB model variants.
  - Benchmarks update and prediction performance, including timing and per-observation metrics.
  - Supports configurable sweeps, repeat counts, output CSV files, model-family selection, and a reduced `--quick` mode.

- **Automation**
  - Added continuous benchmark workflow for pull requests and manual runs, with downloadable results.

- **Tests**
  - Added comprehensive coverage for benchmark configuration, validation, timing, output, and end-to-end execution.

- **Release**
  - Updated package version to 8.1.0.


Whatsonyourmind:
Pushed `1cf3beb` addressing the automated review — three of its points were real, and one of them mattered enough to change the numbers.

**Repeats were sharing state.** `update` mutates the model, but warm-up and all three timed repeats ran against the same instance, so repeats two and three were timing an already-fitted model being retrained on data it had already seen — not the cold update the row claims to report. `time_call` now takes a factory and builds a fresh model per iteration *outside* the timed region. `predict` doesn't mutate, so it still reuses one instance; rebuilding there would only add cold-start cost without changing what is measured.

**The step budget was over-counted.** I had `steps_per_update(chunk) * n_calls`, which charges every action for the whole batch. An update fits one network per action present, and each network sees only its own rows, so the figure is now summed over per-action row counts and multiplied by the objective count for the MO variants. With two actions and a 20-step budget it reports 40 rather than 20.

**`--repeats 0` produced `NaN`.** An empty timing list after the warm-up went straight into the CSV. Rejected at the argument boundary now, and `time_call` guards it too.

Three regression tests cover these — a stateful fake asserting each repeat sees `updates == 1`, the zero-repeats guard, and per-action step accounting including the MO multiplier. 14 tests, ~4s.

One note on the ruff `B023` warnings it raised: those came from a newer ruff than the one pinned in `.pre-commit-config.yaml` (0.11.0), so they wouldn't have failed CI. They were fair, though — the loop variables are now bound explicitly, and I checked the file against both 0.11.0 and 0.15.21 with format and check clean on each.

shaharbar1:
Hey @Whatsonyourmind, sorry for the inconvenience, but we are quite strict on commit and pr description format, you can look at the corresponding md file or previous PRs.

benchmarks/benchmark_models.py
```diff
@@ -0,0 +1,646 @@
...
+#: still dominating one-off overhead.
+FIXED_NUM_STEPS: int = 20
+
+#: ``epochs``/``batch_size`` for the ``epochs`` cMAB pass. The batch size is what makes the
+#: step count scale with n; see the module docstring.
+EPOCHS: int = 2
+EPOCH_BATCH_SIZE: int = 128
+
+
+@dataclass
```
shaharbar1: You can inherit pybanditsbasemodel, it inherits in turn from pydantic basemodel, so it can be used instead of a data class...

pybandits/benchmark_models.py
shaharbar1, line 1: I think that it can be put under the pybandits folder so it is a part of the releases...

benchmarks/benchmark_models.py
```diff
@@ -0,0 +1,646 @@
...
+import csv
+import os
+import sys
+import time
+from dataclasses import dataclass, fields
+from typing import Callable, Dict, List, Optional, Sequence, Tuple
+
+# tqdm honours this natively; the SVI training loop draws a progress bar per update, which
+# would otherwise emit thousands of lines during a sweep. Set before pybandits is imported.
+os.environ.setdefault("TQDM_DISABLE", "1")
```
shaharbar1: You can use the old way:
from tqdm import tqdm
from functools import partialmethod
tqdm.__init__ = partialmethod(tqdm.__init__, disable=True)

This way no need for all the sequential #noqa

tests/test_benchmark_models.py
```diff
@@ -0,0 +1,236 @@
...
+import argparse
+import csv
+import importlib.util
+import sys
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+_BENCHMARK_PATH = Path(__file__).resolve().parents[1] / "benchmarks" / "benchmark_models.py"
```
shaharbar1: Use tmpdir native pytest fixture

tests/test_benchmark_models.py
```diff
@@ -0,0 +1,236 @@
...
+    assert not any(spec.contextual for spec in benchmark_models.build_smab_specs())
+
+
+@pytest.mark.parametrize("multi_objective", [False, True])
+def test_reward_shape_matches_the_model(multi_objective):
+    """Multi-objective models take one reward per objective; the others take a scalar."""
+    spec = benchmark_models.ModelSpec("m", lambda **kw: None, contextual=False, multi_objective=multi_objective)
+    rewards = benchmark_models.make_rewards(spec, n_samples=4, rng=np.random.default_rng(0))
+
+    assert len(rewards) == 4
```
shaharbar1: Please avoid hard coded constants on test as much as possible. Make 4 a test  argument, or use st.given to randomize it. Note there is a dedicated test skill on the repo...

.github/workflows/continuous_benchmark.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Install project dependencies with Poetry
+        run: |
+          poetry install --all-extras
+
+      # Pull requests run the reduced sweep so the SVI path does not dominate CI time; it
+      # exercises the identical code path, so a break in the benchmark still surfaces here.
+      - name: Run benchmark
+        run: |
+          if [ "${{ inputs.quick }}" = "false" ] && [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
```
shaharbar1, line 52: What is the difference between quick and not quick?

.github/workflows/continuous_benchmark.yml
shaharbar1, line 1: Maybe call it continuous_benchmark, like most of the yml workflows?

shaharbar1 (changes requested):
You should also bump the minor version on pyproject.toml.

Whatsonyourmind:
Thanks — and no inconvenience, I should have matched the merged PRs rather than following CONTRIBUTING literally. Fixed in `8150a2d`:

**Format.** The commit subject and PR title are now a plain imperative sentence matching what's on `develop` (`Add limited actions exploration throttle`, `Add per-update decay_factor …`), and the description follows `.github/pull_request_template.md`. I had used the `pybandits-XXXX short_description` form that `CONTRIBUTING.md` still documents — worth knowing that file has drifted from the enforced conventions in a couple of places; it also says branch from `master` (gone) and that the style gate is Flake8, where `.pre-commit-config.yaml` actually runs ruff. Happy to send a small PR correcting it if that's useful.

**Version.** Bumped the minor to `8.1.0`.

**One thing I can't do myself:** the `check conventional commits labels` job requires one of `bug` / `documentation` / `enhancement` / `skip-changelog` on the PR, and I don't have label permissions on the repo — `AddLabelsToLabelable` is refused for my account. Could you add `enhancement`? That check can't go green without it. Everything else that has reported is passing, including the new `benchmark` job.

Also folded in the two remaining automated-review points, both fair: the benchmark workflow's checkout now sets `persist-credentials: false` (nothing in it pushes), and the step-accounting test registers both arms in its fake so the sum genuinely covers two per-action partitions instead of reading one network twice.

14 tests, ruff format and check clean.

shaharbar1:
> Thanks — and no inconvenience, I should have matched the merged PRs rather than following CONTRIBUTING literally. Fixed in `8150a2d`:
> 
> **Format.** The commit subject and PR title are now a plain imperative sentence matching what's on `develop` (`Add limited actions exploration throttle`, `Add per-update decay_factor …`), and the description follows `.github/pull_request_template.md`. I had used the `pybandits-XXXX short_description` form that `CONTRIBUTING.md` still documents — worth knowing that file has drifted from the enforced conventions in a couple of places; it also says branch from `master` (gone) and that the style gate is Flake8, where `.pre-commit-config.yaml` actually runs ruff. Happy to send a small PR correcting it if that's useful.
> 
> **Version.** Bumped the minor to `8.1.0`.
> 
> **One thing I can't do myself:** the `check conventional commits labels` job requires one of `bug` / `documentation` / `enhancement` / `skip-changelog` on the PR, and I don't have label permissions on the repo — `AddLabelsToLabelable` is refused for my account. Could you add `enhancement`? That check can't go green without it. Everything else that has reported is passing, including the new `benchmark` job.
> 
> Also folded in the two remaining automated-review points, both fair: the benchmark workflow's checkout now sets `persist-credentials: false` (nothing in it pushes), and the step-accounting test registers both arms in its fake so the sum genuinely covers two per-action partitions instead of reading one network twice.
> 
> 14 tests, ruff format and check clean.

Thanks!
You can fix contributing.md on this pr or a new one, that's a minor change. Thanks for raising it. We'll check the fixed pr tomorrow.

Whatsonyourmind:
Thanks for merging #157.

One thing worth flagging about the re-run here, because it failed for a reason that isn't the fix: re-running an existing run replays the workflow file from the original attempt, so it pulled the old action again.

```
Download action repository 'danielchabr/pr-labels-checker@v3.3' (SHA:7145ecb...)
##[error]No commit found for SHA: feature/pybandits-78-benchmark-script
```

That run was created on 07-28 and this was attempt 2, so it never saw the step now on `develop` — same failure, same cause as before. The `mheap` step only applies to a **new** `pull_request` event on this branch.

Anything that fires a fresh event should clear it: a push, a close/reopen, or toggling a label off and on. I'm happy to rebase this branch onto the new `develop` so the check runs clean — there's no approval on this PR to reset, so it costs nothing — but I didn't want to move the branch under you unasked. Say the word and I'll push, or toggle the label if you'd rather leave it where it is.

shaharbar1:
> Thanks for merging #157.
> 
> One thing worth flagging about the re-run here, because it failed for a reason that isn't the fix: re-running an existing run replays the workflow file from the original attempt, so it pulled the old action again.
> 
> ```
> Download action repository 'danielchabr/pr-labels-checker@v3.3' (SHA:7145ecb...)
> ##[error]No commit found for SHA: feature/pybandits-78-benchmark-script
> ```
> 
> That run was created on 07-28 and this was attempt 2, so it never saw the step now on `develop` — same failure, same cause as before. The `mheap` step only applies to a **new** `pull_request` event on this branch.
> 
> Anything that fires a fresh event should clear it: a push, a close/reopen, or toggling a label off and on. I'm happy to rebase this branch onto the new `develop` so the check runs clean — there's no approval on this PR to reset, so it costs nothing — but I didn't want to move the branch under you unasked. Say the word and I'll push, or toggle the label if you'd rather leave it where it is.

Yes, please, go ahead and rebase.

Whatsonyourmind:
Pushed as `b3f89ba` — rebased onto `develop`, still one commit, `ruff format --check` and `ruff check` clean and `pytest tests/test_benchmark_models.py -q` green locally (14 passed).

One thing you'll need to do before it means anything: the runs are all sitting at `action_required`. The force-push reset the fork approval gate, so nothing has executed yet — including the label check, which is now absent rather than red. So the fix from #157 is still unproven on a fork PR until you approve-and-run.

Separately, CodeRabbit's nitpick on the rebased head is valid — I checked it against the code rather than taking it at face value. `workflow_dispatch` defaults `quick` to `false`, and the full sweep pairs `update_sizes` up to `10_000` with `update_batches` including `1`, so a manual dispatch can do on the order of 10k sequential single-row SVI fits per cMAB model, pass and repeat. Pull requests are unaffected — `--quick` pins `update_batches=(100,)` and `update_sizes=(100,)`. I'd fix it either by saying so in the `--update-batches` help text or by capping the incremental batch sweep for cMABs specifically. I've deliberately kept it out of this push rather than force-push a second time and reset the approval gate again; tell me which you'd prefer and I'll fold it in whenever you next want the branch moved.

shaharbar1:
> Pushed as `b3f89ba` — rebased onto `develop`, still one commit, `ruff format --check` and `ruff check` clean and `pytest tests/test_benchmark_models.py -q` green locally (14 passed).
> 
> One thing you'll need to do before it means anything: the runs are all sitting at `action_required`. The force-push reset the fork approval gate, so nothing has executed yet — including the label check, which is now absent rather than red. So the fix from #157 is still unproven on a fork PR until you approve-and-run.
> 
> Separately, CodeRabbit's nitpick on the rebased head is valid — I checked it against the code rather than taking it at face value. `workflow_dispatch` defaults `quick` to `false`, and the full sweep pairs `update_sizes` up to `10_000` with `update_batches` including `1`, so a manual dispatch can do on the order of 10k sequential single-row SVI fits per cMAB model, pass and repeat. Pull requests are unaffected — `--quick` pins `update_batches=(100,)` and `update_sizes=(100,)`. I'd fix it either by saying so in the `--update-batches` help text or by capping the incremental batch sweep for cMABs specifically. I've deliberately kept it out of this push rather than force-push a second time and reset the approval gate again; tell me which you'd prefer and I'll fold it in whenever you next want the branch moved.

All the actions completed successfully. Did you finish going over my comments so I can go over the PR again?

Whatsonyourmind:
Yes — all seven are addressed now. Listing them rather than just saying "done", since a couple involved a judgement call:

**1. `Result` and `ModelSpec` inherit `PyBanditsBaseModel`** instead of `@dataclass`. One knock-on worth flagging: pydantic doesn't accept positional arguments, so the ten spec constructions moved to keyword form. `Result` is built after `time_call` returns rather than inside the timed region, so validation cost doesn't land in any measurement.

**2. Moved to `pybandits/benchmark_models.py`** so it ships with the package. Two things followed from that: the workflow now runs `python -m pybandits.benchmark_models`, and the test dropped its `importlib.util.spec_from_file_location` loader for a plain `from pybandits import benchmark_models`. `tqdm` was already a main dependency — `offline_policy_evaluator` imports it — so nothing new enters the runtime requirements by importing it from a shipped module.

**3. tqdm — your version, and it does work below the imports.** I checked rather than assumed, because the reason for the `# noqa: E402` was that `TQDM_DISABLE` had to be set *before* pybandits was imported, and the patch would have been no improvement if it carried the same ordering constraint. It doesn't: patching `__init__` binds at construction time, not import time. With `TQDM_DISABLE` cleared from the environment, a single 20-row `CmabBernoulli.update` gives 143,104 characters of stderr and 4,026 bar glyphs unpatched, and 0 of each with the patch sitting below the import block. All three `# noqa: E402` are gone.

**4. `tmp_path`** — already in use for the end-to-end CSV test.

**5. Test constants** — the `4`, the seed, and the three cMAB pass values are module-level constants passed as default arguments, per the pytest-writer skill.

**6. quick vs not quick.** `--quick` pins `update_sizes=(100,)`, `predict_sizes=(1, 100)`, `action_counts=(2,)`, `update_batches=(100,)` and `repeats=1`; the full sweep runs `update_sizes` up to 10000, `update_batches` including 1, two action counts and three repeats. That is in the `parse_args` docstring now instead of only in my head.

That also answers CodeRabbit's nitpick the cheaper of the two ways: the `--update-batches` help text now spells out that a batch size of 1 means one model fit per row, so the full sweep's 10000×1 cell is 10000 sequential SVI fits per cMAB model per repeat. If you would rather cap the incremental sweep for cMABs than document it, say so and I will swap it.

**7. `continuous_benchmark.yml`** — renamed the file to match the other `continuous_*` workflows. I deliberately left `name: Benchmark` as it was: that string is the check name, so changing it would want branch protection updated to match. Say the word if you would like the display name changed too.

Locally: `ruff format --check` and `ruff check` clean, `pytest tests/test_benchmark_models.py` 14 passed, and `python -m pybandits.benchmark_models --quick --family smab` writes the expected 20-row CSV.

One practical note: this push resets the fork approval gate again, so the runs will need another approve from you before they mean anything.

Whatsonyourmind:
CodeRabbit raised two things on that push. One is real and is now fixed in `d6562b7`; the other I've skipped, with a reason.

**The tqdm patch was leaking process-wide — valid, and fixed.** Patching `tqdm.__init__` at module scope mutates the class for the whole process, so simply *importing* the module silenced progress bars for unrelated code — including every other test in a pytest session, since `tests/test_benchmark_models.py` imports it at collection time. Measured: a plain `tqdm.tqdm(range(3))` writes 88 characters of stderr, and 0 after `from pybandits import benchmark_models`. That only became reachable by users because the module now lives inside the package, so it is worth closing.

It is now a `suppressed_progress_bars()` context manager wrapping `run(args)` inside `main()`, which restores the original `__init__` on exit. Verified all three directions: after import 88 (bars work again), inside the block 0, after the block exits 88. And the CLI path still does the job — `main(['--quick','--family','cmab'])` end to end gives 0 bar glyphs and no `it/s`, writing the expected 41-row CSV.

One correction worth flagging, since it would otherwise look like the obvious fix: **CodeRabbit's suggested remedy — reverting to `os.environ.setdefault("TQDM_DISABLE", "1")` — does not fix the problem it correctly identified.** The env var has exactly the same process-wide reach; I measured it at 0 characters too, identical to the module-level monkeypatch. It also asked me to confirm `TQDM_DISABLE` works on the pinned tqdm — it does — but that is beside the point, because working is precisely what makes it leak. The scoping is what matters, not the mechanism.

**Skipped: the `permissions:` / `timeout-minutes` suggestion on the workflow.** No workflow in this repository sets either — including `continuous_integration.yml`, which also uses `upload-artifact`. Adding a least-privilege block to just the one file I'm introducing would make it the odd one out, and restricting `GITHUB_TOKEN` scope looks like a repo-wide policy call for you rather than something to slip in through a benchmark PR. Happy to add both if you want them, either here or across the workflows.

`ruff format --check` and `ruff check` clean, `pytest tests/test_benchmark_models.py` 14 passed.

This push re-gates the runs again — though they were still sitting unapproved from the last one, so it shouldn't cost you an extra approval.

pybandits/benchmark_models.py
```diff
@@ -0,0 +1,678 @@
...
+    time rather than import time, which is what lets the suppression live here instead of
+    forcing the imports out of order.
+
+    It is scoped to a context manager rather than applied at module level because this
+    module now ships inside the package: a module-level patch would mutate ``tqdm`` for the
+    whole process on import, silencing progress bars for unrelated library code — including
+    every other test in a pytest session that collects this module. Setting ``TQDM_DISABLE``
+    at import instead has exactly the same process-wide reach, so it is not an alternative.
+    """
+    original = tqdm.__init__
```
shaharbar1, line 97: Nice and compact!

pybandits/benchmark_models.py
```diff
@@ -0,0 +1,678 @@
...
+        ModelSpec(
+            name="CmabBernoulliCC",
+            factory=lambda n_actions, **kw: CmabBernoulliCC.cold_start(action_ids_cost=_action_costs(n_actions), **kw),
+            contextual=True,
+            multi_objective=False,
+        ),
+        ModelSpec(
+            name="CmabBernoulliMO",
+            factory=lambda n_actions, **kw: CmabBernoulliMO.cold_start(
+                action_ids=set(_action_ids(n_actions)), n_objectives=DEFAULT_N_OBJECTIVES, **kw
```
shaharbar1: Change kw to kwargs across the script for consistency with rest of the repo.

.github/workflows/continuous_benchmark.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: actions/checkout@v4
+        with:
+          # Nothing here pushes; not retaining the token in the git config keeps
+          # it out of reach of anything the benchmark run executes.
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
```
shaharbar1: Pls make it 3.12.

pybandits/benchmark_models.py
```diff
@@ -0,0 +1,678 @@
...
+    timings = []
+    for _ in range(repeats):
+        call = make_call()  # fresh model, constructed outside the timed region
+        start = time.perf_counter()
+        call()
+        timings.append(time.perf_counter() - start)
+    return float(np.median(timings))
+
+
+def steps_per_update(model: object, n_samples: int) -> int:
```
shaharbar1: Use poitiveint from pydantic, rather than int

pybandits/benchmark_models.py
```diff
@@ -0,0 +1,678 @@
...
+    """
+    results: List[Result] = []
+    for n_actions in action_counts:
+        rng = np.random.default_rng(seed)
+        model = spec.factory(n_actions=n_actions, random_seed=seed, **model_kwargs)
+        for n_samples in sizes:
+            if spec.contextual:
+                context = rng.normal(size=(n_samples, n_features))
+
+                def make_call(model=model, context=context):
```
shaharbar1: you can define make_call with **kwargs for both smab and cmab and pass it on into predict.

pybandits/benchmark_models.py
```diff
@@ -0,0 +1,678 @@
...
+            contextual=False,
+            multi_objective=True,
+        ),
+    ]
+
+
+def build_cmab_specs() -> List[ModelSpec]:
+    """Every contextual Bernoulli MAB exposed by the library."""
+    return [
+        ModelSpec(
```
shaharbar1, line 283: Please also add for CmabBernoulli a config with a hidden layer of let's say 256 neurons.

pybandits/benchmark_models.py
```diff
@@ -0,0 +1,678 @@
...
+        default=EPOCH_BATCH_SIZE,
+        help="Batch size for the 'epochs' cMAB pass; this is what makes steps scale with n.",
+    )
+    parser.add_argument("--output", default=None, help="CSV output path. Defaults to stdout.")
+    parser.add_argument("--quick", action="store_true", help="Small sweeps, for CI.")
+
+    args = parser.parse_args(argv)
+    if args.quick:
+        args.update_sizes = (100,)
+        args.predict_sizes = (1, 100)
```
shaharbar1: I'm not a fan of all of this hard coding. My suggestion is to define to dicts at the top, one for full mode defaults, and the other for quick mode defaults.

tests/test_benchmark_models.py
shaharbar1, line 1: In general, the tests can be highly improved for better mainability using hypothesis (top priority) or pytest parameterize (secondery priority). You could try using the dedicated commited task skill.

shaharbar1 (changes requested):

Whatsonyourmind:
All seven addressed and pushed as `eadd4ee`, rebased onto `develop` at `903865c`, still one commit. Tests went 14 cases to 50, and the file got faster rather than slower: 37.2s to 8.4s. `ruff format --check` and `ruff check` both clean on ruff 0.11.13, matching the `v0.11.0` pre-commit pin rather than a 0.15.x that would have reformatted things CI never asked for.

**1. `kw` to `kwargs`.** All of them, not just the one you pointed at: the ten `ModelSpec` factory lambdas in `build_smab_specs` and `build_cmab_specs`, the splats inside the MO/MOCC bodies, and three stubs in the test file. `grep -n '\bkw\b'` now returns nothing. Ruff pushed a few of those lambdas onto multiple lines because the longer name crossed 120 chars, so some of that diff is the formatter, not me.

**2. Python 3.12** in `continuous_benchmark.yml`. Confirmed it sits inside `python = ">=3.9,<3.15"` and is already in `continuous_delivery.yml`'s matrix.

**3. `PositiveInt`.** `n_samples`, `n_actions`, `n_calls` on `Result`, plus the sweep parameters, `time_call`, `cmab_passes` and both parser hooks.

Two fields are `NonNegativeInt` rather than `PositiveInt`, because zero is meaningful for them and I would rather flag it than quietly widen: `batch_size`, where 0 is the sentinel for "one call carrying all rows", and `steps`, which is 0 for the closed-form sMABs that have no SVI budget. The docstring now says why.

While doing this I found `--seed` was still `type=int` at the parser while `benchmark_predict` and `benchmark_update` both declare `seed: NonNegativeInt`, so `--seed -5` got through argparse and only failed further in. Added a `_non_negative_int` hook to match, with two tests. That one is mine, not yours, so say if you would rather it went in separately.

**4. Single `make_call`.** One `def make_call(model, **kwargs)` at the top of `benchmark_predict`; the loop builds `predict_kwargs` as `{"context": ...}` for cMABs or `{"n_samples": ...}` for sMABs and passes it through. `time_call` wants a zero-argument factory, so the kwargs bind with `functools.partial` rather than changing its contract. Nice side effect: the `model=model, context=context` default-argument trick the old closures needed to avoid late-binding is gone.

**5. `CmabBernoulliHidden256`.** New spec right after the default `CmabBernoulli`, wired through `hidden_dim_list=[256]`. Checked it actually reaches the network rather than landing in an ignored dict: `BaseMab.cold_start` forwards `**kwargs` to `cls(...)`, and `BayesianNeuralNetwork.cold_start` takes `hidden_dim_list: Optional[List[PositiveInt]]`. `build_cmab_specs()` now returns `['CmabBernoulli', 'CmabBernoulliHidden256', 'CmabBernoulliBAI', 'CmabBernoulliCC', 'CmabBernoulliMO', 'CmabBernoulliMOCC']`.

One knock-on: spec `name` is no longer 1:1 with a class name, so `test_every_model_family_is_registered` now asserts every class name appears among the specs plus name-uniqueness, instead of an exact literal set. It still catches a dropped variant, it just tolerates extra configurations of one class.

**6. Two defaults tables.** `FULL_DEFAULTS` and `QUICK_DEFAULTS` at the top, keyed by argparse dest. The four `DEFAULT_*` sweep constants and the inline `default=3` are gone, every parser default reads from `FULL_DEFAULTS`, and the `if args.quick:` block is now a loop over `QUICK_DEFAULTS`.

I gave both tables identical keys so `--quick` reads as a literal diff, which also answers your earlier question about what quick actually changes. That meant leaving the mode-independent scalars (`DEFAULT_N_FEATURES`, `DEFAULT_SEED`, `EPOCHS` and friends) as plain constants rather than duplicating them into both tables. A test asserts the two key sets stay equal.

**7. Tests.** Rewritten against `.claude/skills/pytest-writer/SKILL.md`, which I had missed on the first pass. 13 functions to 23, 13 `@given` and 13 `@parametrize` where there had been no hypothesis at all. `tmp_path` replaces the manual temp handling. The literal 4 is now `st.integers(MIN_SAMPLES, MAX_SAMPLES)`, and the other magic values followed: action counts, repeats, step budgets, rows per step, epochs, sweep values. Constants are injected as default arguments per rule 1 rather than read from module globals inside the bodies.

One deliberate call, since the skill and the surrounding code disagree. The skill says twice not to add `@settings` at all, that deadline and `max_examples` belong to the engineer rather than the test author. About twenty existing tests in `test_actions_manager.py` and `test_cmab.py` do use `@settings(deadline=None)`. I followed the skill, and kept the model-touching cases out of hypothesis so nothing needs the escape hatch: `--hypothesis-show-statistics` reports every `@given` test hitting 100 examples at roughly 0 to 3ms each against the 200ms default. If you would rather they match the legacy files, that is a one-line change per test.

I also checked the properties are load-bearing rather than vacuous: breaking `total_update_steps` so it charges every action for the whole chunk fails both parametrizations of the test that covers it, and dropping the `_non_negative_int` guard fails its rejection test. Restored byte-identical after each.

**Version.** Bumped to `8.2.0`, and the reason is worth flagging because the naive check is misleading. `develop` moved to `8.1.0` via #155 after this branch was cut, so `git diff upstream/develop HEAD -- pyproject.toml` was *empty*: the bump you asked for on 07-28 had silently stopped being one. That is also what the conflict on this rebase was. It is 8.2.0 now, and it will need re-reconciling if something else merges to `develop` before this does.

Full suite still collects clean, 886 tests, no import breakage anywhere.

Same fork caveat as last time: the force-push has reset the approval gate, so the runs are sitting at `action_required` until you approve-and-run. Nothing has executed on `eadd4ee` yet.

